PR #4, opened 2020-05-27
Define #memo_wise and #reset_memo_wise

jemmaissroff:
This PR defines `MemoWise#memo_wise` to memoize methods and `MemoWise#reset_memo_wise` to undo that memoization of methods.

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -3,6 +3,46 @@
 require "memo_wise/version"
 
 module MemoWise
-  class Error < StandardError; end
-  # Your code goes here...
+  def initialize(*values)
+    @_memo_wise = {}
+    super
+  end
+
+  def self.prepended(target)
+    class << target
+      # Implements memoization for the given method name.
+      #
+      # @param method_name [Symbol]
+      #   Name of method for which to implement memoization.
+      def memo_wise(method_name)
+        unless method_name.is_a?(Symbol)
+          raise ArgumentError,
+                "#{method_name.inspect} must be a Symbol"
+        end
+
+        method_visibility = if private_method_defined?(method_name)
+                              :private
+                            elsif protected_method_defined?(method_name)
+                              :protected
+                            else
+                              :public
+                            end
+
+        not_memoized_name = :"_not_memoized_#{method_name}"
+        alias_method not_memoized_name, method_name
+        private not_memoized_name
+
+        module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
+          def #{method_name}(*args)
+            key = [:#{method_name}, args].freeze
+            @_memo_wise.fetch(key) do
+              @_memo_wise[key] = #{not_memoized_name}(*args)
+            end
+          end
+
+          #{method_visibility} :#{method_name}
+        END_OF_METHOD
+      end
+    end
+  end
 end
--- a/spec/memo_wise_spec.rb
+++ b/spec/memo_wise_spec.rb
@@ -1,7 +1,142 @@
 # frozen_string_literal: true
 
 RSpec.describe MemoWise do
-  it "has a version number" do
-    expect(MemoWise::VERSION).not_to be nil
+  let(:class_with_memo) do
+    Class.new do
+      prepend MemoWise
+
+      def initialize
+        @no_args_counter = 0
+        @with_positional_args_counter = 0
+        @with_keyword_args_counter = 0
+        @special_chars_counter = 0
+        @private_memowise_method_counter = 0
+        @protected_memowise_method_counter = 0
+        @public_memowise_method_counter = 0
+      end
+
+      attr_reader :no_args_counter,
+                  :with_positional_args_counter,
+                  :with_keyword_args_counter,
+                  :special_chars_counter,
+                  :private_memowise_method_counter,
+                  :protected_memowise_method_counter,
+                  :public_memowise_method_counter
+
+      def no_args
+        @no_args_counter += 1
+        "no_args"
+      end
+      memo_wise :no_args
+
+      def with_positional_args(a, b) # rubocop:disable Naming/MethodParameterName
+        @with_positional_args_counter += 1
+        "with_positional_args: a=#{a}, b=#{b}"
+      end
+      memo_wise :with_positional_args
+
+      def with_keyword_args(a:, b:) # rubocop:disable Naming/MethodParameterName
+        @with_keyword_args_counter += 1
+        "with_keyword_args: a=#{a}, b=#{b}"
+      end
+      memo_wise :with_keyword_args
+
+      def special_chars?
+        @special_chars_counter += 1
+        "special_chars?"
+      end
+      memo_wise :special_chars?
+
+      def private_memowise_method
+        @private_memowise_method_counter += 1
+        "private_memowise_method"
+      end
+      memo_wise :private_memowise_method
+      private :private_memowise_method
+
+      def protected_memowise_method
+        @protected_memowise_method_counter += 1
+        "protected_memowise_method"
+      end
+      memo_wise :protected_memowise_method
+      protected :protected_memowise_method
+
+      def public_memowise_method
+        @public_memowise_method_counter += 1
+        "public_memowise_method"
+      end
+      memo_wise :public_memowise_method
+      public :public_memowise_method
+    end
+  end
+
+  let(:instance) { class_with_memo.new }
+
+  describe "#memo_wise" do
+    it "memoizes methods with no arguments" do
+      expect(Array.new(4) { instance.send(:no_args) }).to all eq("no_args")
+      expect(instance.no_args_counter).to eq(1)
+    end
+
+    it "memoizes methods with positional arguments" do
+      expect(Array.new(4) { instance.with_positional_args(1, 2) }).
+        to all eq("with_positional_args: a=1, b=2")
+
+      expect(Array.new(4) { instance.with_positional_args(1, 3) }).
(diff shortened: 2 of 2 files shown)


lib/memo_wise.rb
```diff
@@ -3,6 +3,49 @@
...
+  end
+
+  def self.prepended(target)
+    class << target
+      # Implements memoization for the given method name.
+      #
+      # The method must be a *pure function* of the value attributes, and
+      # therefore referentially transparent. Replaces the method with a new
+      # method of the same name, which memoizes the result of calling the
+      # original method. See module comment above for a usage example.
```
ms-ati, line 18: Perhaps this comment should change (could be in a later PR, just calling it out) as the concept of referential transparency, when I wrote this original comment, was specific to the narrow concern of **adding memoization to a Value object**. 

Given expressed goal for this gem to target memoization in general, not limited to value objects, I think this comment is no longer useful?

lib/memo_wise.rb
```diff
@@ -3,6 +3,49 @@
...
+        not_memoized_name = "#{method_name}_not_memoized".to_sym
+        alias_method not_memoized_name, method_name
+        private not_memoized_name
+
+        define_method(method_name) do |*args|
+          key = [method_name, args].freeze
+          @_memo.fetch(key) do
+            @_memo[key] = send(not_memoized_name, *args)
+          end
+        end
```
ms-ati, line 32: For speed, we will need to replace the meta-programming implementation with the code generation implementation, right?

lib/memo_wise.rb
```diff
@@ -3,6 +3,49 @@
...
+      #
+      # @param method_name [Symbol]
+      #   Name of method for which to undo memoization.
+      def reset_memo_wise(method_name)
+        if instance_methods.include?(method_name)
+          not_memoized_name = "#{method_name}_not_memoized".to_sym
+
+          public not_memoized_name
+          alias_method method_name, not_memoized_name
+          undef_method not_memoized_name
```
ms-ati, line 46: Can we have spec coverage and code to make sure what happens if you call `reset_memo_wise` on a name that you did not previously call `memo_wise` on?

jemmaissroff: I was trying to get at this with the context `"when the method being reset is not defined"`. Are you seeing something else I could add there?

lib/memo_wise.rb
```diff
@@ -3,6 +3,49 @@
...
+          end
+        end
+      end
+
+      # Undoes memoization for the given method name. If the method was
+      # previously memoized, it will no longer be memoized.
+      #
+      # @param method_name [Symbol]
+      #   Name of method for which to undo memoization.
+      def reset_memo_wise(method_name)
```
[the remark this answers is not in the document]
jemmaissroff: I'm not sure how to think of this. I buy the argument that we don't use it, but I'm unsure how to know or weigh if others do or would use it. Any thoughts there?

lib/memo_wise.rb
ms-ati: I think I missed it! Thanks @jemmaissroff

lib/memo_wise.rb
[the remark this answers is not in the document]
ms-ati: @JacobEvelyn I think we should split-track your two points: (a) do we need the `reset_memo_wise` in this gem, given that Panorama doesn't currently use it, and (b) separately, do we need either an API or just documentation on how to pre-set a memoized value, or to un-set or re-set an alread-memoized value, because we do those things at Panorama by hacking the instance variable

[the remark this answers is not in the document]
ms-ati: My perspective on _(a) do we need the reset_memo_wise in this gem, given that Panorama doesn't currently use it?_:

I think we **do need this in a public gem**, because I see this as expected API surface for a competitive memoization library in the public sphere to have -- even if Panorama code doesn't currently use it. In terms of trade-offs, I don't currently perceive the size of this code+specs, nor the cognitive load of the complexity of this feature in the API, to be large enough to trade-off against having this feature. In terms of Ruby DSLs in general, I find having a way to reset or undo changes to a class to be uniformly expected in the wider community, particularly with regard to methods being aliased. Thoughts?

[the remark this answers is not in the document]
ms-ati: My current perspective on _(b) separately, do we need either an API or just documentation on how to pre-set a memoized value, or to un-set or re-set an already-memoized value, because we do those things at Panorama by hacking the instance variable currently_:

I think we **do need this feature in a public API**, because it's useful enough that we depend on hacking around it in Panorama code. But I'd be happy seeing it come in a subsequent PR after this initial one is merged. Thoughts?

lib/memo_wise.rb
```diff
@@ -3,6 +3,46 @@
 require "memo_wise/version"
 
 module MemoWise
-  class Error < StandardError; end
-  # Your code goes here...
+  def initialize(*values)
```
ms-ati, line 6: Can we have YARD doc of the `initialize` method, including an `@example` section of how it works after you `prepend` to a class and then call `.new`?

lib/memo_wise.rb
```diff
@@ -3,6 +3,46 @@
...
+      def memo_wise(method_name)
+        not_memoized_name = "#{method_name}_not_memoized".to_sym
+        alias_method not_memoized_name, method_name
+        private not_memoized_name
+
+        module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
+          def #{method_name}(*args)
+            key = ["#{method_name}", args].freeze
+            @_memo_wise.fetch(key) do
+              @_memo_wise[key] = send(:"#{not_memoized_name}", *args)
```
ms-ati, line 26: We can wait till benchmarking later if you want, but I know already that it will be measurably faster to interpolate the actual method call into this string, instead of meta-programming via `send` -- do we want to change this now?

lib/memo_wise.rb
```diff
@@ -3,6 +3,46 @@
...
+    super
+  end
+
+  def self.prepended(target)
+    class << target
+      # Implements memoization for the given method name.
+      #
+      # @param method_name [Symbol]
+      #   Name of method for which to implement memoization.
+      def memo_wise(method_name)
```
ms-ati, line 17: How would you feel about raising `ArgumentError` if `method_name` is NOT a `Symbol`, and adding specs? I see benefits in:
1. I can suggest some optimizations knowing for sure we have a Symbol
2. We have a documented and tested API error case surface, to add and update over time

lib/memo_wise.rb
```diff
@@ -3,6 +3,46 @@
...
+        not_memoized_name = "#{method_name}_not_memoized".to_sym
+        alias_method not_memoized_name, method_name
+        private not_memoized_name
+
+        module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
+          def #{method_name}(*args)
+            key = ["#{method_name}", args].freeze
```
ms-ati, line 24: Why convert `method_name` to String here? I am very confident it will be measurably faster for this key lookup comparison to be a tuple of (Symbol, Array) than it would be for (String, Array)?

lib/memo_wise.rb
```diff
@@ -3,6 +3,46 @@
...
+              @_memo_wise[key] = send(:"#{not_memoized_name}", *args)
+            end
+          end
+        END_OF_METHOD
+      end
+
+      # Undoes memoization for the given method name. If the method was
+      # previously memoized, it will no longer be memoized.
+      #
+      # @param method_name [Symbol]
```
ms-ati, line 35: How would you feel about also documenting the expected `NoMethodError` if the given method was not previously memoized?  This would be the `@raise` tag in YARD doc ([see here](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#raise)).

I see documenting expected errors as part of a well-maintained API documentation, so I look for parity between all three of:
1. What the code can/will raise in the implementation
2. What the YARD docs say can be raised
3. What the RSpec coverage tests will be raised

lib/memo_wise.rb
ms-ati: Ok @jemmaissroff I'm closing this comment as you are correct it's already covered by a spec, but I added a new comment asking if we can document the exception raising as part of the YARD doc elsewhere

lib/memo_wise.rb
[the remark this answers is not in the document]
JacobEvelyn: I'm happy to be overridden here, and I hear what you're saying @ms-ati. Thanks for the thoughts!

> (a) do we need the reset_memo_wise method in this gem, given that Panorama doesn't currently use it?

I guess I don't see the harm in leaving this out until either we or another gem user needs it. While I do see this functionality in other gems like `Memoist`, a large API footprint in those gems has actually been something that discourages me from using them.

> (b) separately, do we need either an API or just documentation on how to pre-set a memoized value, or to un-set or re-set an already-memoized value, because we do those things at Panorama by hacking the instance variable currently?

If we can come up with a reasonable API that we ourselves would use, I'm totally down to add it. (And I agree with you that we can wait for a subsequent PR to add it.)

lib/memo_wise.rb
jemmaissroff: Switched it to be a symbol here, but it needs to be interpolated because it's within the larger `module_eval` string

lib/memo_wise.rb
jemmaissroff: Deferring to a future PR!

lib/memo_wise.rb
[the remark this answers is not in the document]
ms-ati: (moved to a later PR)

PR #13, opened 2020-07-14
Add initial benchmarks

JacobEvelyn:


benchmarks/.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.1
```
JacobEvelyn, line 1: Ultimately it might be nice to have benchmarks run in multiple versions of Ruby, but for the time being it was easiest for me to just hardcode this to the latest.

benchmarks/Gemfile
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+ruby "2.7.1" # Keep updated to latest.
+
+gem "benchmark-ips"
```
JacobEvelyn, line 7: I've used both `benchmark` (part of stdlib) and `benchmark-ips` and I'm not sure if `benchmark-ips` is still the Ruby best practice. I went with it mostly because it makes a nicer report.

Thoughts?

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+# using it to minimize the chance that our benchmarks are affected by ordering.
+BENCHMARK_GEMS = [
+  BenchmarkGem.new(Memery, :include, :memoize),
+  BenchmarkGem.new(MemoWise, :prepend, :memo_wise),
+  BenchmarkGem.new(Memoist, :extend, :memoize),
+  BenchmarkGem.new(Memoized, :include, :memoize),
+  BenchmarkGem.new(Memoizer, :include, :memoize)
+].shuffle
+
+# Use metaprogramming to ensure that each class is created in exactly the
```
JacobEvelyn, line 23: I don't love doing so much metaprogramming here, but without this the file got really big and I worried about subtle errors (e.g. accidentally copy-pasting something wrong and testing the wrong gem). Thoughts?

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+
+# We benchmark different cases separately, to ensure that slow performance in
+# one method or code path isn't hidden by fast performance in another.
+Benchmark.ips do |x|
+  BENCHMARK_GEMS.each do |benchmark_gem|
+    example_class = Object.const_get("#{benchmark_gem.klass}Example")
+    x.report("#{benchmark_gem.klass}#no_args") do |n_times|
+      n_times.times do
+        instance = example_class.new
+        10_000.times { instance.no_args }
```
JacobEvelyn, line 54: I chose `10_000` arbitrarily. Is there a better way I should choose?

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+    def keyword_args(a:, b:)
+      300 + a + b
+    end
+    send(benchmark_gem.memoization_method, :keyword_args)
+  end)
+end
+
+# We benchmark different cases separately, to ensure that slow performance in
+# one method or code path isn't hidden by fast performance in another.
+Benchmark.ips do |x|
```
JacobEvelyn, line 48: Should we use the default `benchmark-ips` warmup/test times (2s/5s), or something else/longer?

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+    send(benchmark_gem.memoization_method, :positional_args)
+
+    def keyword_args(a:, b:)
+      300 + a + b
+    end
+    send(benchmark_gem.memoization_method, :keyword_args)
+  end)
+end
+
+# We benchmark different cases separately, to ensure that slow performance in
```
JacobEvelyn, line 46: Is there a better way to group these benchmarks?

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+# one method or code path isn't hidden by fast performance in another.
+Benchmark.ips do |x|
+  BENCHMARK_GEMS.each do |benchmark_gem|
```
JacobEvelyn, line 49: Should we disable GC for these tests? See: https://github.com/evanphx/benchmark-ips#custom-suite

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+    example_class = Object.const_get("#{benchmark_gem.klass}Example")
+    x.report("#{benchmark_gem.klass}#no_args") do |n_times|
+      n_times.times do
+        instance = example_class.new
+        10_000.times { instance.no_args }
+      end
+    end
+  end
+
+  x.compare!
```
JacobEvelyn, line 59: Initial results—not great:

```
Comparison:
    Memoized#no_args:      882.6 i/s
     Memoist#no_args:      421.9 i/s - 2.09x  (± 0.00) slower
    Memoizer#no_args:      334.9 i/s - 2.64x  (± 0.00) slower
      Memery#no_args:       85.6 i/s - 10.31x  (± 0.00) slower
    MemoWise#no_args:       75.1 i/s - 11.75x  (± 0.00) slower
```

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+    example_class = Object.const_get("#{benchmark_gem.klass}Example")
+    x.report("#{benchmark_gem.klass}#positional_args") do |n_times|
+      n_times.times do
+        instance = example_class.new
+        10_000.times { instance.positional_args(index, index + 10) }
+      end
+    end
+  end
+
+  x.compare!
```
JacobEvelyn, line 73: Initial results—decent but not amazing:

```
Comparison:
MemoWise#positional_args:       56.4 i/s
Memoized#positional_args:       51.8 i/s - 1.09x  (± 0.00) slower
Memoizer#positional_args:       46.5 i/s - 1.21x  (± 0.00) slower
Memery#positional_args:       46.0 i/s - 1.23x  (± 0.00) slower
Memoist#positional_args:       38.4 i/s - 1.47x  (± 0.00) slower
```

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,88 @@
...
+    example_class = Object.const_get("#{benchmark_gem.klass}Example")
+    x.report("#{benchmark_gem.klass}#keyword_args") do |n_times|
+      n_times.times do
+        instance = example_class.new
+        10_000.times { instance.keyword_args(a: index, b: index + 10) }
+      end
+    end
+  end
+
+  x.compare!
```
JacobEvelyn, line 87: Initial results—good:

```
Comparison:
MemoWise#keyword_args:       34.4 i/s
Memoized#keyword_args:       25.5 i/s - 1.35x  (± 0.00) slower
Memoizer#keyword_args:       24.4 i/s - 1.41x  (± 0.00) slower
 Memery#keyword_args:       24.2 i/s - 1.42x  (± 0.00) slower
Memoist#keyword_args:       21.3 i/s - 1.62x  (± 0.00) slower
```

benchmarks/benchmarks.rb
jemmaissroff: Not sure how to answer this. Can't think of why the default wouldn't / shouldn't work for our use case.

benchmarks/benchmarks.rb
jemmaissroff: Are the results consistent with bigger numbers?

benchmarks/benchmarks.rb
jemmaissroff: Not for this PR, but I wonder if it's worth branching on the no args case?

benchmarks/benchmarks.rb
jemmaissroff: This would be more metaprogramming, but I wonder if it'd be more complex or more straightforward to store a map of proc name --> block to execute and then iterate through that map calling the benchmark functionality over it?

benchmarks/benchmarks.rb
JacobEvelyn: Really I just worry that the way I'm adding metaprogramming will add confounding factors that influence the results. Would appreciate review with an eye toward "yes this seems okay" or "no this will invalidate our results."

benchmarks/Gemfile
JacobEvelyn: Ruby's built-in `benchmark` has results that [look like this](https://ruby-doc.org/stdlib-2.5.0/libdoc/benchmark/rdoc/Benchmark.html):

```
              user     system      total        real
foo:      0.960000   0.000000   0.960000 (  0.957966)
bar:      0.960000   0.000000   0.960000 (  0.960423)
baz:      0.950000   0.000000   0.950000 (  0.954864)
```

i.e. no ordering by speed, no "1.3x slower," and no "±" to indicate standard deviation.

benchmarks/benchmarks.rb
JacobEvelyn: Done!

benchmarks/Gemfile
jemmaissroff: I'm with you on `benchmark-ips` then, specifically I find the ordering by speed and the "1.3x slower" to be very helpful

benchmarks/benchmarks.rb
jemmaissroff: Talked over zoom, I think we addressed any potential confounding factors.

benchmarks/Gemfile
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+# Keep updated to latest Ruby, and keep in sync with benchmarks/.ruby-version.
+ruby "2.7.1"
+
+gem "benchmark-ips"
+gem "memery"
+gem "memo_wise", git: "https://github.com/panorama-ed/memo_wise.git"
```
ms-ati, line 10: Uh oh @JacobEvelyn shouldn’t this source the MemoWise from the relative directory locally? That is, wouldn’t we expect the benchmarks to run against the codebase version they are sitting in, including potentially uncommitted local changes?

benchmarks/benchmarks.rb
```diff
@@ -0,0 +1,120 @@
...
+end
+
+Benchmark.ips do |x|
+  x.config(suite: suite)
+  BENCHMARK_GEMS.each do |benchmark_gem|
+    example_class = Object.const_get("#{benchmark_gem.klass}Example")
+    x.report("#{benchmark_gem.klass}#positional_args") do
+      instance = example_class.new
```
ms-ati, line 97: Could we create the instance outside the report block?

benchmarks/Gemfile
JacobEvelyn: Ah, I forgot we could do that as we do with `outtake`!

benchmarks/benchmarks.rb
JacobEvelyn: Talked over Zoom; we couldn't find a good way to do this.

benchmarks/benchmarks.rb
JacobEvelyn: Yes, good catch! That was left over from when I previously had each test creating many instances. Will change.

JacobEvelyn:
> Not blocking, but in a later PR, I recommend adding some assurances that every test is actually hitting a memorized method (for example, by raising an exception in the methods that are memorized if they are ever called twice on one instance).

Thanks, Marc! I thought about this but worried that the obvious implementations (storing a set of args passed to each method) would be doing similar operations to the gem itself and might muddy the benchmark results. One alternative I'd thought of was to just make the operations something like:

```
def no_args
  sleep 1
  100
end
```

But then that's of course also going to change the benchmark results somewhat. Thoughts?

PR #14, opened 2020-07-16
Optimization for methods with no arguments

jemmaissroff:
Going to wait for benchmarking to merge before making claims on hard numbers here, but checking out the `first-benchmarks` branch, it looks like this doesn't have a negative effect on any methods with arguments, and speeds up the no args case to only ~2x the best gem we're comparing against.

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -33,17 +33,29 @@ module MemoWise
         alias_method not_memoized_name, method_name
         private not_memoized_name
 
-        module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
-          def #{method_name}(*args)
-            key = [:#{method_name}, args].freeze
-            @_memo_wise.fetch(key) do
-              @_memo_wise_keys[:#{method_name}] << args
-              @_memo_wise[key] = #{not_memoized_name}(*args)
+        if instance_method(method_name).arity.zero?
+          module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
+            def #{method_name}
+              @_memo_wise.fetch(:#{method_name}) do
+                @_memo_wise[:#{method_name}] = #{not_memoized_name}
+              end
             end
-          end
+          END_OF_METHOD
+        else
+          module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
+            def #{method_name}(*args)
+              key = [:#{method_name}, args].freeze
+              @_memo_wise.fetch(key) do
+                @_memo_wise_keys[:#{method_name}] << args
+                @_memo_wise[key] = #{not_memoized_name}(*args)
+              end
+            end
+          END_OF_METHOD
+        end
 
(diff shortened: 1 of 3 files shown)


lib/memo_wise.rb
```diff
@@ -34,7 +34,11 @@ def memo_wise(method_name)
 
         module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
           def #{method_name}(*args)
-            key = [:#{method_name}, args].freeze
+            key = if args.empty?
```
JacobEvelyn, line 37: @jemmaissroff how would you feel about pulling this logic outside the method so it's not getting called every time? I'm thinking something like:

```ruby
if method(:method_name).arity.zero?
  method_name_and_args = method_name
  key = ":#{method_name}"
else
  method_name_and_args = "#{method_name}(*args)"
  key = "[:#{method_name}, args].freeze"
end

module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
  def #{method_name_and_args}
    key = #{key}
    ...
```

though I'd also be open to just having separate `module_eval`s based on the different versions.

JacobEvelyn: Also this is the most nitpicky thing but I'm realizing now that `method_name.inspect` may be a hair more efficient than `":#{method_name}"` because it's able to more easily allocate exactly the string size it needs. (Though to be honest I don't have a great sense for everything [the source](https://ruby-doc.org/core-2.7.1/Symbol.html#method-i-inspect) is doing.)

jemmaissroff: This is really neat! And looks like makes us faster than the other gems for the `no_args` case too! Pushed another version here.

lib/memo_wise.rb
```diff
@@ -34,25 +34,29 @@ def memo_wise(method_name)
         private not_memoized_name
 
         if instance_method(method_name).arity.zero?
-          method_name_and_args = "#{method_name}()"
-          key = method_name.inspect
+          module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
+            def #{method_name}
+              key = #{method_name.inspect}
```
JacobEvelyn, line 39: I know I was the one who originally led you to change this to `inspect`, but since we're elsewhere using `:#{method_name}` and this is in the middle of a bigger string interpolation now I'd vote we just change this back for consistency (or change the other instances of `:#{method_name}` to `#{method_name.inspect}`).

lib/memo_wise.rb
```diff
@@ -34,25 +34,29 @@ def memo_wise(method_name)
...
+              end
             end
-          end
+          END_OF_METHOD
+        end
 
-          #{method_visibility} :#{method_name}
-        END_OF_METHOD
+        module_eval <<-END_OF_VISIBILITY, __FILE__, __LINE__ + 1
+          "#{method_visibility} #{method_name.inspect}"
```
JacobEvelyn, line 58: Same here!

PR #15, opened 2020-07-17
Implements #reset_memo_wise

jemmaissroff:
This PR implements resetting of memoization. Once a method has already been memoized, it allows us to reset any memoized values while still leaving the method itself memoized.

Outside of the scope of this PR (but to come in later PR)
- YARDocs to describe this method

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -5,6 +5,7 @@ require "memo_wise/version"
 module MemoWise
   def initialize(*values)
     @_memo_wise = {}
+    @_memo_wise_keys = Hash.new { |h, k| h[k] = [] }
     super
   end
 
@@ -36,6 +37,7 @@ module MemoWise
           def #{method_name}(*args)
             key = [:#{method_name}, args].freeze
             @_memo_wise.fetch(key) do
+              @_memo_wise_keys[:#{method_name}] << args
               @_memo_wise[key] = #{not_memoized_name}(*args)
             end
           end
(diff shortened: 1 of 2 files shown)


lib/memo_wise.rb
```diff
@@ -36,6 +37,7 @@ def memo_wise(method_name)
           def #{method_name}(*args)
             key = [:#{method_name}, args].freeze
             @_memo_wise.fetch(key) do
+              @_memo_wise_keys[:#{method_name}] << args
```
JacobEvelyn, line 40: I'm curious (not saying this isn't the best approach, I really don't know) to learn more about why you went down this route. I can see a few different implementation paths:

1. What you have here.
2. Delete this line. On reset, iterate through all keys and delete the ones that start with `:#{method_name}`.
3. Same as 2, but change keys to be another level of a Hash rather than an array. On reset, just delete the key.

Should we proceed with this as-is and revisit it when we have benchmarks merged? Should we add benchmarks for `reset_memo_wise` at all?

lib/memo_wise.rb
```diff
@@ -45,4 +47,21 @@ def #{method_name}(*args)
       end
     end
   end
+
+  def reset_memo_wise(method_name)
+    unless respond_to?(method_name)
```
JacobEvelyn, line 52: Should we similarly enforce that this is a symbol? Otherwise I think this might not clear correctly if we pass in a string.

lib/memo_wise.rb
jemmaissroff: I was thinking about different paths here and totally agree with your thought about re-visiting when benchmarks are merging. Since this gem is attempting to optimize for speed, that seems like the best criteria with which to make this decision!

I can look too, but out of curiosity, do you remember most other gems having reset methods too?

JacobEvelyn: Yep I think we'd probably want benchmarks for these worst-case scenarios:

- the case where memoized methods are getting called with lots of different arguments
- the case where method results are getting memoized and reset repeatedly

PR #17, opened 2020-07-24
Implements #preset_memo_wise to enable us to preset memoization values

jemmaissroff:
Implements `#preset_memo_wise` with an API that has `instance.preset_memo_wise(method_name, *args) { value }`

To come in future PRs:
- Validate params sent to `#preset_memo_wise`
- YARDocs for `#preset_memo_wise`

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -29,15 +29,15 @@ module MemoWise
                               :public
                             end
 
-        not_memoized_name = :"_not_memoized_#{method_name}"
-        alias_method not_memoized_name, method_name
-        private not_memoized_name
+        original_memo_wised_name = :"_memo_wise_original_#{method_name}"
+        alias_method original_memo_wised_name, method_name
+        private original_memo_wised_name
 
         if instance_method(method_name).arity.zero?
           module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
             def #{method_name}
               @_memo_wise.fetch(:#{method_name}) do
-                @_memo_wise[:#{method_name}] = #{not_memoized_name}
+                @_memo_wise[:#{method_name}] = #{original_memo_wised_name}
               end
             end
           END_OF_METHOD
@@ -47,7 +47,7 @@ module MemoWise
               key = [:#{method_name}, args].freeze
               @_memo_wise.fetch(key) do
                 @_memo_wise_keys[:#{method_name}] << args
-                @_memo_wise[key] = #{not_memoized_name}(*args)
+                @_memo_wise[key] = #{original_memo_wised_name}(*args)
               end
             end
           END_OF_METHOD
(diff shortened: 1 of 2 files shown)


lib/memo_wise.rb
```diff
@@ -81,4 +81,28 @@ def reset_all_memo_wise
     @_memo_wise_keys.clear
     @_memo_wise.clear
   end
+
+  # If no block is given, will set the value to nil
```
jemmaissroff, line 85: YARDocs to follow in a later PR

lib/memo_wise.rb
```diff
@@ -81,4 +81,28 @@ def reset_all_memo_wise
...
+  def preset_memo_wise(method_name, *args)
+    not_memoized_name = :"_not_memoized_#{method_name}"
+    unless self.class.private_method_defined?(not_memoized_name)
```
jemmaissroff, line 88: Thoughts on this approach to checking if a method has been memoized? Another option I was considering is just storing a set of memoized methods and checking for presence of `method_name` in that set.

lib/memo_wise.rb
JacobEvelyn, line 85: > If no block is given, will set the value to nil

What's the thought behind that when we already have `reset_memo_wise`?

lib/memo_wise.rb
```diff
@@ -81,4 +81,28 @@ def reset_all_memo_wise
...
+      raise ArgumentError, "#{method_name} is not a memoized method"
+    end
+
+    validate_params!(not_memoized_name, args.dup)
```
JacobEvelyn, line 92: I'm hesitant to use `.dup` until we know we need to. What am I missing?

spec/memo_wise_spec.rb
```diff
@@ -329,4 +341,146 @@ def public_memowise_method
...
+    it "does not preset memoization methods across instances" do
+      instance2 = class_with_memo.new
+
+      instance.preset_memo_wise(:no_args) { "preset_no_args" }
+
+      expect(instance2.no_args).to eq("no_args")
+      expect(instance2.no_args_counter).to eq(1)
+    end
+
+    context "when the method to preset memoization for is not memoized" do
```
JacobEvelyn, line 472: Should we also add tests for `reset_memo_wise` for this case?

spec/memo_wise_spec.rb
```diff
@@ -329,4 +341,146 @@ def public_memowise_method
...
+      expect(instance.true_method_counter).to eq(0)
+    end
+
+    it "presets memoization for methods set to nil values" do
+      instance.preset_memo_wise(:no_args) { nil }
+      expect(Array.new(4) { instance.no_args }).to all eq(nil)
+      expect(instance.no_args_counter).to eq(0)
+    end
+
+    context "when there is already a memoized value" do
```
JacobEvelyn, line 398: Since these are basically the same tests as the above (except for the first and last lines), I'm wondering if there's an easy way to use a `shared_examples` block for both of these. Thoughts?

spec/memo_wise_spec.rb
JacobEvelyn: I am. I don't think I see a reason why these would differ. Do you?

spec/memo_wise_spec.rb
```diff
@@ -343,123 +343,103 @@ def unmemoized_method
...
+        it "presets memoization" do
+          instance.preset_memo_wise(:no_args) { nil }
+          expect(Array.new(4) { instance.no_args }).to all eq(nil)
+          expect(instance.no_args_counter).to eq(expected_counter)
+        end
       end
     end
 
+    it_behaves_like "presets memoization", overriding: false
+    it_behaves_like "presets memoization", overriding: true
```
JacobEvelyn, line 441: Slight preference for giving these meaningful contexts:

```
context "when memoized values were not already set" do
  it_behaves_like "presets memoization", overriding: false
end

context "when memoized values were already set" do
  it_behaves_like "presets memoization", overriding: true
end
```

PR #19, opened 2020-07-28
Various tweaks to benchmarking

JacobEvelyn:
This commit:

- Changes benchmarks to print the gem versions
- Includes all dependency versions in the benchmark results
  in the `README.md`
- Changes our GitHub Actions to run benchmarks
- Pins benchmark dependency versions in the `Gemfile`


benchmarks/.bundle/config
```diff
@@ -0,0 +1,3 @@
+---
+BUNDLE_RETRY: "3"
```
jemmaissroff, line 2: Where are these used?

JacobEvelyn: Honestly I didn't realize this file existed or was committed. Cautiously removing!

.github/workflows/main.yml
```diff
@@ -21,7 +19,9 @@ jobs:
         key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
         restore-keys: |
           ${{ runner.os }}-gems-
-    - run: bundle config path vendor/bundle
```
ms-ati, line 24: Why was this removed? This is part of the bundle caching setup isn't it?

benchmarks/benchmarks.rb
```diff
@@ -6,7 +6,9 @@
 require "memo_wise"
 require "memoist"
 require "memoized"
+require "memoized/version" # Memoizer::VERSION does not get loaded above.
```
ms-ati, line 9: Commend should say Memoize**d**, yes?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,3 @@
+**Before merging:**
+
+- [ ] Copy the latest benchmark results into the `README.md` and update this PR
```
ms-ati, line 3: Should this say "if they have changed beyond the expected slight noise"? Or similar?

Should this checklist item link to the Github Action results area or guide the contributor to know where to look for the results?

Can we guide them more specifically to `README.md` in the section *Section Name* or something?

.github/workflows/main.yml
```diff
@@ -6,10 +6,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ 2.5, 2.6, 2.7, jruby ]
+        ruby: [jruby, 2.5, 2.6, 2.7]
```
ms-ati, line 9: No problem with changing the order here, but was this on purpose and if so what is the reasoning behind the order?

.dependabot/config.yml
```diff
@@ -1,29 +1,13 @@
 # Reference: https://dependabot.com/docs/config-file/
 version: 1
 update_configs:
-  # This configures dependency updates for one package manager. In some
```
JacobEvelyn, line 4: These comments didn't seem useful, and were somewhat confusing for external readers.

.dependabot/config.yml
```diff
@@ -1,29 +1,13 @@
...
-  # projects, such as warehouse, where we have Ruby and Python, there can be
-  # separate package_manager entries.
   - package_manager: "ruby:bundler"
     directory: "/"
     update_schedule: "weekly"
-
-    default_labels:
```
JacobEvelyn, line 11: I don't think we want `Needs QA`, and `dependencies` is the default.

.dependabot/config.yml
```diff
@@ -1,29 +1,13 @@
...
-
     automerged_updates:
-      # This allows all dependencies that are used for development, e.g., rspec,
-      # rspec-mock, vcr, etc, to be automatically updated. This is generally
-      # okay because the dependencies are not used in production.
       - match:
           dependency_type: "development"
           update_type: "all"
+- package_manager: "ruby:bundler"
+    directory: "/benchmarks"
```
JacobEvelyn, line 12: This allows us to get Dependabot updates for benchmark gems too.

.dependabot/config.yml
```diff
@@ -1,29 +1,13 @@
...
-
-    # Dependabot will use a repository's default branch. This will override
-    # that.
-    # target_branch: "master"
-
-    allowed_updates:
-      - match:
-          dependency_type: "direct"
-
+    update_schedule: "live"
```
JacobEvelyn, line 6: I don't see an advantage to doing these weekly, since the main reason for that was to isolate our Buildkite Dependabot builds but this repo doesn't use Buildkite. Open to pushback though!

.github/workflows/main.yml
JacobEvelyn: This was on purpose; I find it easier to think about since the latest Ruby version is also the last item in the list and the one that we run linters/benchmarks on.

.github/workflows/main.yml
JacobEvelyn: I was playing around with using the built-in [`bundler-cache: true` option](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically) in the `ruby/setup-ruby` action, which would let us remove a bunch of code here. Unfortunately, when I enable that the `jruby` build stops working and I wasn't able to figure out why. (I did make https://github.com/ruby/setup-ruby/issues/72)

.github/PULL_REQUEST_TEMPLATE.md
JacobEvelyn: > Should this say "if they have changed beyond the expected slight noise"? Or similar?

I'm torn. I agree it may not be that useful but I worry about us missing changes because we're used to not changing this.

> Should this checklist item link to the Github Action results area or guide the contributor to know where to look for the results?

Sure. I'm not sure how to link to the results since I think the PR number will be in the URL. Do you know of a way?

> Can we guide them more specifically to README.md in the section Section Name or something?

Sure.

README.md
```diff
@@ -24,13 +24,16 @@ TODO: Write usage instructions here
...
+[`benchmark-ips`](https://github.com/evanphx/benchmark-ips) 2.8.2:
+
 |Gem|Method with no arguments|Method with positional arguments|Method with keyword arguments|
 |---|------------------------|--------------------------------|-----------------------------|
-|**`memo_wise`**|**baseline**|**baseline**|**baseline**|
-|`memery`|13.02x (± 0.00) slower|1.27x (± 0.00) slower|1.49x (± 0.00) slower|
-|`memoist`|2.65x (± 0.00) slower|1.46x (± 0.00) slower|1.57x (± 0.00) slower|
-|`memoized`|1.27x  (± 0.00) slower|1.16x (± 0.00) slower|1.37x (± 0.00) slower|
-|`memoizer`|3.18x (± 0.00) slower|1.25x (± 0.00) slower|1.45x (± 0.00) slower|
+|**`memo_wise` (0.1.0)**|**baseline**|**baseline**|**baseline**|
```
JacobEvelyn, line 32: These are copied from an earlier GitHub Actions run for this PR.

.github/workflows/main.yml
```diff
@@ -6,10 +6,8 @@ jobs:
...
     runs-on: ubuntu-latest
-    env:
```
JacobEvelyn, line 11: I couldn't figure out what this was for, so I removed it.

.github/PULL_REQUEST_TEMPLATE.md
jemmaissroff: > I'm torn. I agree it may not be that useful but I worry about us missing changes because we're used to not changing this.

I actually like the idea of updating it each time too. I think it gives complete objectivity to our benchmarks as opposed to introducing any subjectivity or potential confusion for any external folks hoping to contribute to this gem who might not know what "expected noise" would be.

ms-ati: Ok! I'm convinced by both reasonings

PR #23, opened 2020-07-31
Reduce potential noise from benchmarks

JacobEvelyn:
This commit changes the benchmarks to solely test memoized
value retrieval. All other costs (instance initialization,
memoized value storage) are removed from benchmarking to
limit ourselves to just what we care about optimizing the
most.

**Before merging:**

### DIFF
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -34,8 +34,8 @@ end
 suite = BenchmarkSuiteWithoutGC.new
 
 BenchmarkGem = Struct.new(:klass, :inheritance_method, :memoization_method) do
-  def version
-    klass::VERSION
+  def benchmark_name
+    "#{klass} (#{klass::VERSION})"
   end
 end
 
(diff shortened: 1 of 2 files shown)


ms-ati:
@JacobEvelyn I'm seeing these errors or warnings in the benchmarks output?

```
MemoWise (0.1.0): positional_args
/home/runner/work/memo_wise/memo_wise/benchmarks/vendor/bundle/ruby/2.7.0/gems/memoizer-1.0.3/lib/memoizer.rb:42: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

JacobEvelyn:
> @JacobEvelyn I'm seeing these errors or warnings in the benchmarks output?
> 
> ```
> MemoWise (0.1.0): positional_args
> /home/runner/work/memo_wise/memo_wise/benchmarks/vendor/bundle/ruby/2.7.0/gems/memoizer-1.0.3/lib/memoizer.rb:42: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
> ```

@ms-ati yep, we've seen these since the beginning (and other gems get them too). I'd noticed that `memery` used the `ruby2_keywords` gem to silence these but hadn't dug further than that. Thanks for reminding me to add it to our todo list to discuss!

benchmarks/benchmarks.rb
```diff
@@ -74,22 +74,24 @@ def keyword_args(a:, b:)
...
-      instance = example_class.new
-      N_INTERATIONS_WITHOUT_ARGUMENTS.times { instance.no_args }
-    end
+    instance = Object.const_get("#{benchmark_gem.klass}Example").new
+
+    # Run once to memoize the result value, so our benchmark only tests memoized
+    # retrieval time.
+    instance.no_args
+
+    x.report("#{benchmark_gem.benchmark_name}: no_args") { instance.no_args }
```
JacobEvelyn, line 94: I mentioned this in my original benchmarks PR, but I'm not sure what to do about the fact that the `benchmark-ips` gem has a vague recommendation that very fast blocks might be better off using the `do |times|` construction: https://github.com/evanphx/benchmark-ips#synopsis

I couldn't find a lot of details on that, but my guess is that the overhead it's referring to is block execution overhead and that using a while loop might be lighter-weight. It seems a little hacky and makes me uneasy though, and I'm pretty sure more recent Ruby versions have optimized block overhead more so I thought it made sense to just use the simpler block version.

ms-ati: I think it's fine to start with this, as it is showing clear distinctions already -- we could always experiment with changing it later

ms-ati: (comment on the PR) I think this is fine, as it's already revealing significant performance differences -- we can always try out the other form later and see if you like it then

PR #25, opened 2020-08-10
Change internal data structure to nested hash

JacobEvelyn:
**Before merging:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,10 @@ Memoized value retrieval time using Ruby 2.7.1 and
 |Gem|Method with no arguments|Method with positional arguments|Method with keyword arguments|
 |---|------------------------|--------------------------------|-----------------------------|
 |**`memo_wise` (0.1.0)**|**baseline**|**baseline**|**baseline**|
-|`memery` (1.3.0)|12.22x (± 0.00) slower|1.28x (± 0.00) slower|1.48x (± 0.00) slower|
-|`memoist` (0.16.2)|2.50x (± 0.00) slower|1.44x (± 0.00) slower|1.61x (± 0.00) slower|
-|`memoized` (1.0.2)|1.21x  (± 0.00) slower|1.17x (± 0.00) slower|1.40x (± 0.00) slower|
-|`memoizer` (1.0.3)|3.34x (± 0.00) slower|1.29x (± 0.00) slower|1.48x (± 0.00) slower|
+|`memery` (1.3.0)|11.80x (± 0.00) slower|2.00x (± 0.00) slower|1.93x (± 0.00) slower|
+|`memoist` (0.16.2)|2.55x (± 0.00) slower|2.20x (± 0.00) slower|2.01x (± 0.00) slower|
+|`memoized` (1.0.2)|1.19x  (± 0.00) slower|1.81x (± 0.00) slower|1.81x (± 0.00) slower|
+|`memoizer` (1.0.3)|2.93x (± 0.00) slower|1.97x (± 0.00) slower|1.90x (± 0.00) slower|
 
 You can run benchmarks yourself with:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,10 +30,10 @@ Memoized value retrieval time using Ruby 2.7.1 and
 |Gem|Method with no arguments|Method with positional arguments|Method with keyword arguments|
 |---|------------------------|--------------------------------|-----------------------------|
 |**`memo_wise` (0.1.0)**|**baseline**|**baseline**|**baseline**|
-|`memery` (1.3.0)|12.22x (± 0.00) slower|1.28x (± 0.00) slower|1.48x (± 0.00) slower|
-|`memoist` (0.16.2)|2.50x (± 0.00) slower|1.44x (± 0.00) slower|1.61x (± 0.00) slower|
-|`memoized` (1.0.2)|1.21x  (± 0.00) slower|1.17x (± 0.00) slower|1.40x (± 0.00) slower|
-|`memoizer` (1.0.3)|3.34x (± 0.00) slower|1.29x (± 0.00) slower|1.48x (± 0.00) slower|
+|`memery` (1.3.0)|11.80x (± 0.00) slower|2.00x (± 0.00) slower|1.93x (± 0.00) slower|
```
JacobEvelyn, line 33: We seem to have gotten slightly slower for the `no_args` case and slightly faster for the other cases. I'm not sure how we want to think about that. Thoughts?

lib/memo_wise.rb
```diff
@@ -4,8 +4,7 @@
 
 module MemoWise
   def initialize(*values)
-    @_memo_wise = {}
-    @_memo_wise_keys = Hash.new { |h, k| h[k] = [] }
+    @_memo_wise = Hash.new { |h, k| h[k] = {} }
```
JacobEvelyn, line 7: I also tried a version where this had no default proc and we just created the sub-hash when necessary, but I didn't notice a performance difference and this is a bit easier for the args cases.

lib/memo_wise.rb
```diff
@@ -70,19 +70,13 @@ def reset_memo_wise(method_name, *args)
...
-      @_memo_wise_keys[method_name].each do |method_args|
-        @_memo_wise.delete([method_name, method_args])
-      end
+      @_memo_wise.delete(method_name)
     else
-      @_memo_wise.delete([method_name, args])
+      @_memo_wise[method_name].delete(args)
     end
-
-    @_memo_wise.delete(method_name)
```
JacobEvelyn, line 80: To be honest I'm not sure why we had this line before; I think it was a typo.

lib/memo_wise.rb
ms-ati: What if, as an experiment, we don't allocate the `{}` in the `Hash` default proc, so that it doesn't impact the no-args cases?

lib/memo_wise.rb
```diff
@@ -42,12 +41,13 @@ def #{method_name}
...
         else
+          # Note that we don't need to freeze args before using it as a hash key
+          # because Ruby always copies argument arrays when splatted.
           module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
             def #{method_name}(*args)
-              key = [:#{method_name}, args].freeze
-              @_memo_wise.fetch(key) do
-                @_memo_wise_keys[:#{method_name}] << args
-                @_memo_wise[key] = #{not_memoized_name}(*args)
+              hash = @_memo_wise[:#{method_name}]
```
ms-ati, line 48:
```suggestion
              hash = (@_memo_wise[:#{method_name}] ||= {})
```

What if we tried allocating a new Hash here?

JacobEvelyn: Wouldn't we still need to assign that hash to `@_memo_wise[:#{method_name}]`? My implementation looked something like:

```
@_memo_wise[:#{method_name}] = {} unless @_memo_wise.key?(:#{method_name})
hash = @_memo_wise[:#{method_name}]
...
```

PR #30, opened 2020-08-26
Preset memo wise

jemmaissroff:
Implements #preset_memo_wise with an API that has instance.preset_memo_wise(method_name, *args) { value }

To come in future PRs:

- Validate params sent to #preset_memo_wise
- YARDocs for #preset_memo_wise

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -2,7 +2,7 @@
 
 require "memo_wise/version"
 
-module MemoWise
+module MemoWise # rubocop:disable Metrics/ModuleLength
   def initialize(*values)
     @_memo_wise = Hash.new { |h, k| h[k] = {} }
     super
@@ -40,9 +40,9 @@ module MemoWise
                               :public
                             end
 
-        not_memoized_name = :"_not_memoized_#{method_name}"
-        alias_method not_memoized_name, method_name
-        private not_memoized_name
+        original_memo_wised_name = :"_memo_wise_original_#{method_name}"
+        alias_method original_memo_wised_name, method_name
+        private original_memo_wised_name
 
         method = instance_method(method_name)
 
@@ -52,7 +52,7 @@ module MemoWise
           module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
             def #{method_name}
               @_memo_wise.fetch(:#{method_name}) do
-                @_memo_wise[:#{method_name}] = #{not_memoized_name}
+                @_memo_wise[:#{method_name}] = #{original_memo_wised_name}
               end
             end
           END_OF_METHOD
@@ -81,7 +81,7 @@ module MemoWise
             def #{method_name}#{args_str}
               hash = @_memo_wise[:#{method_name}]
               hash.fetch(#{fetch_key}) do
-                hash[#{fetch_key}] = #{not_memoized_name}#{args_str}
+                hash[#{fetch_key}] = #{original_memo_wised_name}#{args_str}
               end
             end
           END_OF_METHOD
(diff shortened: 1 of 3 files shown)


lib/memo_wise.rb
```diff
@@ -106,21 +108,55 @@ def reset_memo_wise(method_name, *args, **kwargs)
...
+
+    unless block
+      raise ArgumentError,
+            "Pass a block as the value to preset for #{method_name}, #{args}"
+    end
+
+    validate_params!(method_name, args)
+
+    if method(method_name).arity.zero?
+      @_memo_wise[method_name] = block_given? ? yield : nil
```
JacobEvelyn, line 130: Why do we have these `block_given?` checks if we enforce there's always a block? Or am I misunderstanding what this code is doing?

lib/memo_wise.rb
```diff
@@ -106,21 +108,55 @@ def reset_memo_wise(method_name, *args, **kwargs)
...
+      raise ArgumentError, "#{method_name} is not a memo_wised method"
+    end
+  end
+
+  def fetch_key(method_name, *args, **kwargs)
+    method = self.class.instance_method(method_name)
+    has_arg = MemoWise.has_arg?(method)
+
+    if has_arg && MemoWise.has_kwarg?(method)
+      [args, kwargs]
```
JacobEvelyn, line 152: In the code above we call `.freeze` on this compound key. Should we add that here or remove that there? (I don't know.)

spec/memo_wise_spec.rb
```diff
@@ -364,7 +376,7 @@ def public_memowise_method
       expect(instance.special_chars_counter).to eq(2)
     end
 
-    it "resets memoization for methods with false values" do
+    it "resets memoization for methods set to false values" do
```
JacobEvelyn, line 379: Should we make the same name change to the test below?

spec/memo_wise_spec.rb
```diff
@@ -400,6 +412,13 @@ def public_memowise_method
       }
     end
 
+    context "when the method to preset memoization for is not memoized" do
```
JacobEvelyn, line 415: Typo: `preset` -> `reset`

spec/memo_wise_spec.rb
```diff
@@ -400,6 +412,13 @@ def public_memowise_method
...
+      it {
```
JacobEvelyn, line 416: I have a slight preference for these multiline blocks to be `do...end`s instead of `{...}`. How would you feel about changing them (here and below)?

JacobEvelyn (changes requested):

spec/memo_wise_spec.rb
ms-ati: Same preference here! I'm surprised that rubocop-rspec doesn't lint that?

ms-ati: Also, are these `it` blocks not named because RSpec magic creates a name from the expectation?

lib/memo_wise.rb
```diff
@@ -106,21 +108,55 @@ def reset_memo_wise(method_name, *args, **kwargs)
...
   end
 
   def reset_all_memo_wise
     @_memo_wise.clear
   end
+
+  def preset_memo_wise(method_name, *args, **kwargs, &block)
+    validate_memo_wised!(method_name)
+
+    unless block
```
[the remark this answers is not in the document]
JacobEvelyn: Still wondering about this!

JacobEvelyn:
Still just the one comment about whether we need `unless block`!

JacobEvelyn (approved):

PR #32, opened 2020-09-08
Separate *args from **kwargs

jemmaissroff:
This commit separates *args from **kwargs in the memo_wise'd
method signatures, and only uses the minimum necessary. This
allows us to gain performance benefits (fewer allocations in
many cases) and silence the Ruby warnings.

**Before merging:**

- [x] Rebase PR into one coherent commit

### DIFF
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -70,6 +70,26 @@ BENCHMARK_GEMS.each do |benchmark_gem|
         100
       end
       #{benchmark_gem.memoization_method} :keyword_args
+
+      def positional_and_keyword_args(a, b:)
+        100
+      end
+      #{benchmark_gem.memoization_method} :positional_and_keyword_args
+
+      def positional_and_splat_args(a, *args)
+        100
+      end
+      #{benchmark_gem.memoization_method} :positional_and_splat_args
+
+      def keyword_and_double_splat_args(a:, **kwargs)
+        100
+      end
+      #{benchmark_gem.memoization_method} :keyword_and_double_splat_args
+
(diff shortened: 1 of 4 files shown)


JacobEvelyn:
@jemmaissroff I hope I didn't step on your toes in finishing this out—I think when we last left off we said I'd update the benchmarks in the README but to do that I realized I needed to add a few more benchmarks and then it was easy to just also add the RSpec tests as well.

This is ready for review!

benchmarks/benchmarks.rb
```diff
@@ -70,6 +70,26 @@ def keyword_args(a:, b:)
         100
       end
       #{benchmark_gem.memoization_method} :keyword_args
+
+      def positional_and_keyword_args(a, b:)
+        100
```
jemmaissroff, line 75: Do we have an issue to fill these in?

jemmaissroff (commented):
Code lgtm. Only question is if we're planning on filling in the benchmarks in this PR or another one?

benchmarks/benchmarks.rb
JacobEvelyn: What would we fill these in with? I don't think the result would impact the benchmark since we're only benchmarking cache value returns, right?

jemmaissroff: Ah yup you're totally right, PR lgtm. Can merge this if you approve!

Also, looks like I lost permission to push to this repo. Any chance you could check on that?

JacobEvelyn: Whoops, sorry! Should be fixed now!

JacobEvelyn (approved):

Issue #34, opened 2020-09-16
Add YARD documentation


JacobEvelyn:
- [x] Create the README badge+link, and setup automatic generation on RubyDoc.info for both Github project and (later) published gem
- [x] Instrument the project’s development dependencies so that we can generate docs exactly like RubyDoc.info locally
- [x] Backfill docs so that they look good on RubyDoc.info
- [x] Backfill docs for meta-programmed methods as well
- [x] Take a pass to ensure the YARD linking (to classes and methods) is working
- [x] Ensure code snippets are rendered with Ruby highlighting
- [x] Try to add test coverage of documentation example code snippets, e.g. with https://github.com/p0deje/yard-doctest

ms-ati:
Initial documentation is done!

PR #52, opened 2020-10-05
Backfill initial yard docs

ms-ati:
First pass through the MemoWise module. Includes all of the non-metaprogrammed class and instance methods. Excludes all of the meta-programmed methods, which will come in a separate PR after we figure out the best way to document them.

We add examples in the YARD doc comments which are tested using the
`yard doctest` plugin.

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -2,24 +2,113 @@
 
 require "memo_wise/version"
 
+# MemoWise is the wise choice for memoization in Ruby.
 module MemoWise # rubocop:disable Metrics/ModuleLength
-  def initialize(*values)
+  # Constructor to setup memoization state before
+  # [calling the original](https://medium.com/@jeremy_96642/ruby-method-auditing-using-module-prepend-4f4e69aacd95)
+  # constructor.
+  #
+  # - **Q:** Why is [Module#prepend](https://ruby-doc.org/core-2.7.1/Module.html#method-i-prepend)
+  #          important here
+  #          ([more info](https://medium.com/@leo_hetsch/ruby-modules-include-vs-prepend-vs-extend-f09837a5b073))?
+  # - **A:** To setup *mutable state* inside the instance, even if the original
+  #          constructor will then call
+  #          [Object#freeze](https://ruby-doc.org/core-2.7.1/Object.html#method-i-freeze).
+  #
+  # This approach supports memoization on frozen (immutable) objects -- for
+  # example, classes created by the
+  # [Values](https://github.com/tcrayford/Values)
+  # [gem](https://rubygems.org/gems/values).
+  #
+  def initialize(*)
     @_memo_wise = Hash.new { |h, k| h[k] = {} }
     super
   end
 
+  # @private
+  #
+  # Determine whether `method` takes any *positional* args.
+  #
+  # These are the types of positional args:
+  #
+  #   * *Required* -- ex: `def foo(a)`
+  #   * *Optional* -- ex: `def foo(b=1)`
+  #   * *Splatted* -- ex: `def foo(*c)`
+  #
+  # @param method [Method, UnboundMethod]
+  #   Arguments of this method will be checked
+  #
+  # @return [Boolean]
+  #   Return `true` if `method` accepts one or more positional arguments
+  #
+  # @example
+  #   class Example
+  #     def no_args
+  #     end
+  #
+  #     def position_arg(a)
+  #     end
+  #   end
+  #
+  #   MemoWise.has_arg?(Example.instance_method(:no_args)) #=> false
+  #
+  #   MemoWise.has_arg?(Example.instance_method(:position_arg)) #=> true
+  #
   def self.has_arg?(method) # rubocop:disable Naming/PredicateName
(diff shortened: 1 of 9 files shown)


lib/memo_wise.rb
```diff
@@ -2,24 +2,109 @@
...
+  #
   def self.has_arg?(method) # rubocop:disable Naming/PredicateName
     method.parameters.any? do |(param, _)|
       param == :req || param == :opt || param == :rest # rubocop:disable Style/MultipleComparison
     end
   end
 
+  # Determine whether `method` takes any *keyword* args.
+  #
+  # These are the types of positional args:
```
ms-ati, line 63:
```suggestion
  # These are the types of keyword args:
```

lib/memo_wise.rb
```diff
@@ -94,42 +179,113 @@ def #{method_name}#{args_str}
...
+      @_memo_wise[method_name] = yield
     else
-      @_memo_wise[method_name].delete(fetch_key(method_name, *args, **kwargs))
+      @_memo_wise[method_name][fetch_key(method_name, *args, **kwargs)] = yield
     end
   end
 
+  # Resets all memoized results -- of all methods.
+  #
+  # @return [void]
```
ms-ati, line 248: Agreed: remove this, and make `method_name` optional on `reset_memo_wise` for this

lib/memo_wise.rb
```diff
@@ -2,24 +2,109 @@
...
+  #     end
+  #
+  #     def position_arg(a)
+  #     end
+  #   end
+  #
+  #   MemoWise.has_arg?(Example.instance_method(:no_args)) #=> false
+  #
+  #   MemoWise.has_arg?(Example.instance_method(:position_arg)) #=> true
+  #
```
ms-ati, line 54: Agreed - make this api private, it shouldn't appear in YARD

lib/memo_wise.rb
```diff
@@ -2,24 +2,109 @@
...
+  #     end
+  #
+  #     def keyword_args(a:, b: 1)
+  #     end
+  #   end
+  #
+  #   MemoWise.has_kwarg?(Example.instance_method(:position_args)) #=> false
+  #
+  #   MemoWise.has_kwarg?(Example.instance_method(:keyword_args)) #=> true
+  #
```
ms-ati, line 87: Agreed - make this api private, it shouldn't appear in YARD

ms-ati:
See #34

lib/memo_wise.rb
```diff
@@ -94,42 +183,164 @@ def #{method_name}#{args_str}
...
   end
 
-  def reset_memo_wise(method_name, *args, **kwargs)
+  # Presets the memoized result for the given method to the result of the given
+  # block.
+  #
+  # This method is for situations where the caller *already* has the result of
+  # an expensive method call, and wants to preset that result as memoized for
+  # future calls. In other words, the memoized method will be called *zero*
+  # times rather than once.
```
jemmaissroff, line 192: Do we want to explicitly name in the docs that we are not validating parameters? That is to say, if a set of invalid parameters is passed to this method, we will *not* raise an error

lib/memo_wise.rb
```diff
@@ -94,42 +183,164 @@ def #{method_name}#{args_str}
...
+  #   ex.method_to_preset #=> "B"
+  #
+  #   ex.method_called_times #=> nil
+  #
+  def preset_memo_wise(method_name, *args, **kwargs)
     validate_memo_wised!(method_name)
 
-    unless method_name.is_a?(Symbol)
-      raise ArgumentError, "#{method_name.inspect} must be a Symbol"
+    unless block_given?
```
jemmaissroff, line 236: Just want to make sure I'm reading this diff right -- it's appearing like this because of the reordering of the methods right, not because there are actual changes to `preset_memo_wise`?

jemmaissroff (changes requested):

lib/memo_wise.rb
ms-ati: Hi @jemmaissroff can you perhaps review the proposed new version of `preset_memo_wise` (that is, rather than the diff with the old version) and see if that answers this question?

lib/memo_wise.rb
ms-ati: @jemmaissroff Sure, I can add that if you'd like

lib/memo_wise.rb
ms-ati: @jemmaissroff is this view helpful, to see the final version of `preset_memo_wise` rather than the diff?

https://github.com/panorama-ed/memo_wise/blob/7aa734d97948cc71330d66fb3c5694f222317d13/lib/memo_wise.rb#L233

lib/memo_wise.rb
ms-ati: Ok, added @jemmaissroff, please review and resolve this thread if it's good?

lib/memo_wise.rb
jemmaissroff: Looks great, thanks for copying it in here

jemmaissroff (approved):

PR #54, opened 2020-10-07
Codecov

jemmaissroff:
Adds code coverage gem to memo_wise.

Still to come in future PR:
- Add badge to memo_wise (Issue #54)
- Set minimum thresholds (Issue #54)
- Fix missing code coverage (Issue #56)

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -14,3 +14,13 @@ RSpec.configure do |config|
     c.syntax = :expect
   end
 end
+
+if ENV["CI"] == "true" && Gem.loaded_specs.key?("codecov")
+  require "codecov"
+  require "simplecov"
+
+  SimpleCov.start do
+    enable_coverage :branch
+  end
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -14,3 +14,13 @@
     c.syntax = :expect
   end
 end
+
+if ENV["CI"] == "true" && /^\d+\.\d+/.match(RUBY_VERSION)[0] == "2.7"
```
ms-ati, line 18: @jemmaissroff in order to keep the Ruby version logic in one place -- just `main.yml` -- what would you think about changing this?

Perhaps the `main.yml` could set an env variable such as "CODECOV"?

But perhaps a better, more robust solution would be to detect whether the `codecov` gem is present for this block? That way, setting the bundle `without` controls whether this block runs?

ms-ati (approved):
LGTM

Issue #56, opened 2020-10-07
Fix Missing Code Coverage


jemmaissroff:
Ensure we're at 100% code coverage for our specs!

Also, mandate 100% branch coverage (see #59 for more context)

Issue #37, opened 2020-09-16
Rework the README and branding


JacobEvelyn:
- [x] update tagline in gemspec and on GitHub: `The wise choice for Ruby memoization`
- [x] add feature list/unique selling points of our gem:
  - Highest performance + Rigorous benchmarking
  - Preset
  - reset
  - Frozen instances (value objects)
  - Test coverage
  - Well-documented API
- [x] Add other GitHub badges (see examples here: https://github.com/JacobEvelyn/friends)

JacobEvelyn:
@jemmaissroff I'm not sure I totally follow the last item in the checklist:

> Document that changing method visibility must occur before memoizing methods in order to preserve visibility. (See #59 for more detail.)

If I do:

```ruby
memo_wise :foo
private :foo
```

or

```ruby
private :foo
memo_wise :foo
```

isn't the result of both of these the same, i.e. `foo` will be both memoized and private?

jemmaissroff:
@JacobEvelyn re:
> isn't the result of both of these the same, i.e. `foo` will be both memoized and private?
Yes, you're actually totally right. I just ran the tests locally switching the order. But can you explain to me how that's working?

My assumption was that if we made the method private after memoizing it, it wouldn't be private when we check `private_method_defined?` within `#memo_wise`. Can you explain to me why that doesn't matter?

JacobEvelyn:
> Can you explain to me why that doesn't matter?

Sure! In this case:

```ruby
memo_wise :foo
private :foo
```

We're first creating a new _public_ method `foo`, which wraps the behavior of the old ([now-private](https://github.com/panorama-ed/memo_wise/blob/main/lib/memo_wise.rb#L150)) method `foo`. But then on the next line when we call `private :foo` we're making our new memoized method private.

Does that make sense?

jemmaissroff:
> Does that make sense?

Ah yes! So just to confirm my understanding. The order of the `private` and `memo_wise` call will impact what we do internally. But in both cases we'll ultimately have a private memoized method

PR #59, opened 2020-10-11
Add CodeCov badge and minimum coverages

jemmaissroff:
This PR:
* Adds CodeCov badge
* Adds `lib/memo_wise.rb` to the files evaluated by CodeCov
* Fixes private / protected memoization to improve CodeCov results
* Fails the test suite when line coverage lowers, or branch coverage falls below 90%

Resolves Issue #59

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -189,9 +189,7 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
           END_OF_METHOD
         end
 
-        module_eval <<-END_OF_VISIBILITY, __FILE__, __LINE__ + 1
-          "#{method_visibility} :#{method_name}"
-        END_OF_VISIBILITY
+        module_eval "#{method_visibility} :#{method_name}", __FILE__, __LINE__
       end
     end
   end
--- a/spec/memo_wise_spec.rb
+++ b/spec/memo_wise_spec.rb
@@ -115,15 +115,15 @@ RSpec.describe MemoWise do
         @private_memowise_method_counter += 1
         "private_memowise_method"
       end
-      memo_wise :private_memowise_method
       private :private_memowise_method
+      memo_wise :private_memowise_method
 
       def protected_memowise_method
         @protected_memowise_method_counter += 1
         "protected_memowise_method"
       end
-      memo_wise :protected_memowise_method
       protected :protected_memowise_method
+      memo_wise :protected_memowise_method
 
       def public_memowise_method
         @public_memowise_method_counter += 1
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,5 +1,30 @@
 # frozen_string_literal: true
 
+# Simplecov needs to be loaded before we require `memo_wise` in order to
+# properly track all memo_wise files
+if Gem.loaded_specs.key?("codecov")
+  require "codecov"
+  require "simplecov"
+
+  SimpleCov.start do
+    enable_coverage :branch
+  end
+
+  SimpleCov.formatter = if ENV["CI"] == "true"
+                          SimpleCov::Formatter::Codecov
+                        else
+                          # Writes coverage file into coverage/index.html
+                          # when run outside of CI for local development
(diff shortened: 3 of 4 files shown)


spec/memo_wise_spec.rb
```diff
@@ -115,15 +115,15 @@ def private_memowise_method
         @private_memowise_method_counter += 1
         "private_memowise_method"
       end
-      memo_wise :private_memowise_method
       private :private_memowise_method
+      memo_wise :private_memowise_method
```
jemmaissroff, line 119: The ordering here meant we were making the methods `private` or `protected` after memoizing them. We need to change their visibility _before_ memoizing them so that our memoization can preserve visibility.

I have added a note to describe this in the README in Issue #37

lib/memo_wise.rb
```diff
@@ -190,7 +190,7 @@ def #{method_name}#{args_str}
         end
 
         module_eval <<-END_OF_VISIBILITY, __FILE__, __LINE__ + 1
-          "#{method_visibility} :#{method_name}"
+          #{method_visibility} :#{method_name}
```
jemmaissroff, line 193: We didn't need the `""` here because the `END_OF_VISIBILITY` tags are wrapping this in a string

JacobEvelyn: Happy to keep this as-is, but would it make sense to just use a regular string instead of a heredoc here?

spec/spec_helper.rb
```diff
@@ -1,5 +1,30 @@
 # frozen_string_literal: true
 
+# Simplecov needs to be loaded before we require `memo_wise` in order to
+# properly track all memo_wise files
+if Gem.loaded_specs.key?("codecov")
```
ms-ati, line 5: Why remove the `ENV[CI] == "true"` check here? This will make it run locally when we run specs?

spec/spec_helper.rb
```diff
@@ -1,5 +1,30 @@
...
+                          SimpleCov::Formatter::Codecov
+                        else
+                          # Writes coverage file into coverage/index.html
+                          # when run outside of CI for local development
+                          SimpleCov::Formatter::HTMLFormatter
+                        end
+
+  # SimpleCov.refuse_coverage_drop is only implemented for line coverage, so for
+  # branch coverage we must use `minimum_coverage`
+  SimpleCov.minimum_coverage branch: 90
```
JacobEvelyn, line 23: Out of curiosity, why 90? What's our branch coverage at now?

spec/spec_helper.rb
jemmaissroff: I actually found it useful to run locally, so moved the `ENV["CI"]` check below to determine the output. When running locally, it now outputs to the `coverage` directory so we can open `coverage/index.html` and clearly view the results. I found this helpful for local dev, but let me know if you're thinking differently!

spec/spec_helper.rb
jemmaissroff: It's at ~93. I figured we could work to improve it and then move this up! The 90 was just arbitrarily lower than what we're currently at. What are you thinking here?

lib/memo_wise.rb
```diff
@@ -189,9 +189,7 @@ def #{method_name}#{args_str}
           END_OF_METHOD
         end
 
-        module_eval <<-END_OF_VISIBILITY, __FILE__, __LINE__ + 1
-          "#{method_visibility} :#{method_name}"
-        END_OF_VISIBILITY
+        module_eval "#{method_visibility} :#{method_name}"
```
JacobEvelyn, line 192: Does RuboCop not still want `, __FILE__, __LINE__ + 1`?

spec/spec_helper.rb
JacobEvelyn: I'd love to see this at 100, so whatever you think is the best way to get there works for me!

jemmaissroff: Yup let's keep at 90 in this PR, and when it's fixed we'll change it to 100. Added this to Issue #56.

JacobEvelyn (approved):

PR #73, opened 2020-10-22
Add support for Class method memoization

ms-ati:
Included in this PR is support for memoization of class methods,
which in Ruby are technically instance methods of the singleton
class.

Support class methods defined in two ways.

(1) The "self." syntax:

    class SelfDot
      prepend MemoWise

      def self.foo; end
      memo_wise :foo
    end

(2) The "class << self" syntax:

    class ClassShiftSelf
      class << self
        prepend MemoWise

        def foo; end
        memo_wise :foo
      end
    end

Planned future PRs:

  (A) Enhance README and Documentation to show class methods

  (B) Refactor the tests into examples that can all be run for all of:
       - instance methods
       - class methods defined first way
       - class methods defined second way

  (C) Refactor the code to extract all meta-programming and reflection
      concerns to utility classes and document and test them independently;
      Leave core 'memo_wise' easier to read

**Before merging:**

- [ ] add a new commit updating the benchmarks with results from GitHub Actions

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -40,7 +40,7 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
   # [gem](https://rubygems.org/gems/values).
   #
   def initialize(*)
-    @_memo_wise = Hash.new { |h, k| h[k] = {} }
+    MemoWise.create_memo_wise_state!(self)
     super
   end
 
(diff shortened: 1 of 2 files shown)


lib/memo_wise.rb
```diff
@@ -114,6 +114,163 @@ def self.has_kwarg?(method) # rubocop:disable Naming/PredicateName
...
+      singleton.private_instance_methods.include?(method_name) ||
+      singleton.instance_methods.include?(method_name)
+
+    if is_singleton_method
+      visibility = method_visibility(singleton, method_name)
+      [:class, visibility, singleton, singleton.instance_method(method_name)]
+    else
+      visibility = method_visibility(klass, method_name)
+      [:instance, visibility, klass, klass.instance_method(method_name)]
+    end
```
ms-ati, line 182: per @jemmaissroff could we simplify the checks to just `singleton.instance_method(method_name)` and same for `klass`? Because it appears like those would simply return `nil` if it's not present, and we already believe that they work for private methods

lib/memo_wise.rb
```diff
@@ -114,6 +114,163 @@ def self.has_kwarg?(method) # rubocop:disable Naming/PredicateName
...
+    raise NameError if name.nil?
+
+    Object.const_get(name) # raises NameError if not found
+  rescue NameError
+    # Option 2: Search ObjectSpace
+    #   * Works in all cases, so makes a good fallback option
+    #   * 1:1 relationship of singleton class to original class is documented
+    #   * Even though this only runs at load time, searches all Class objects
+    ObjectSpace.each_object(Class).find do |cls|
+      !cls.singleton_class? && cls.singleton_class == klass
```
ms-ati, line 249: Per @jemmaissroff can we remove the `singleton_class?` check and just have the 2nd one?

JacobEvelyn (commented):
This is an impressive PR, @ms-ati ! Thanks for taking it on. ❤️ I don't have any significant feedback on the code itself right now.

I worry that I created a lot of avoidable complexity when I suggested we support the `class << self` syntax—I think my main concern was avoiding method name collisions, and it seemed like if other gems handled class methods this way maybe it was the easiest approach. (I actually thought we'd decided to _only_ support `class << self` and _not_ support `def self.foo` but it sounds like I might have been misunderstanding what was being discussed.)

To that end, I'm wondering:

1. Are we sure we want to support both of these ways of memoizing class methods? I don't have strong feelings and am happy to keep this as-is (in one sense it's nice to be able to support users' varied styles), but I wonder if it would simplify our code and docs to just pick one.
2. If we support the `def self.foo` style (either alone or in addition to the other style), are we concerned about method name collisions where both `Example.foo` and `Example#foo` are defined?

ms-ati:
Hi Jake!

1. I definitely think we should support both syntaxes - and I don't think that supporting both method of memoizing class methods is adding a lot of avoidable complexity that could be avoided by supporting only one. Contrarily, I think that there's a fair amount of essential complexity in memoizing Class methods. But I see paths forward for future PRs to reduce the incidental complexity quite a bit, and am excited to get here.

2. In the current code, I think you are correct that only the `self.foo` can be memoized. Should I add failing test case for this configuration and look for a way to fix it? Or should this be a follow-on PR?

```ruby
class Example
  prepend MemoWise

  def self.foo
    456
  end
  memo_wise :foo # wraps the class method

  def foo
    123
  end
  memo_wise :foo # should wrap the instance method?
end
```

lib/memo_wise.rb
```diff
@@ -114,6 +114,163 @@ def self.has_kwarg?(method) # rubocop:disable Naming/PredicateName
...
+    # on the "singleton class" of a given Class object, found via
+    # {Class#singleton_class}.
+    # See: https://medium.com/@leo_hetsch/demystifying-singleton-classes-in-ruby-caf3fa4c9d91
+    singleton = klass.singleton_class
+
+    is_singleton_method =
+      singleton.private_instance_methods.include?(method_name) ||
+      singleton.instance_methods.include?(method_name)
+
+    if is_singleton_method
```
JacobEvelyn, line 176: Would we want to consider making this the `else` case so that when the same name exists for both a class and instance method we default to memoizing the instance one (which I think will be more common)?

lib/memo_wise.rb
```diff
@@ -114,6 +114,163 @@ def self.has_kwarg?(method) # rubocop:disable Naming/PredicateName
...
+    end
+
+    # Option 1: Derive from `#to_s` of the singleton class
+    #   * Works unless original class NOT assigned to constant
+    #   * Depends on an undocumented naming convention in Ruby, where singleton
+    #     class of `Foo` will return "#<Class:Foo>" from `#to_s` method.
+    name = klass.to_s.match(/#<Class:([^>]+)>/)&.[](1)
+    raise NameError if name.nil?
+
+    Object.const_get(name) # raises NameError if not found
```
JacobEvelyn, line 242: Is there a risk of this finding something but it not being the class we expect? We could mitigate this with:

```
Object.const_get(name).tap do |original|
  raise NameError unless original.singleton_class == klass
end
```

or just remove the `Option 1` approach entirely? Especially if this runs at load time I wonder if it's really worth it to optimize this so much, but I could be convinced either way.

JacobEvelyn:
> 1. I definitely think we should support both syntaxes for defining and memoizing class methods, as they are both widely used, and I think we'll see both in Panorama's code as well.

That's fine by me, though I actually don't think Panorama would use the `class << self` for memoization if we could avoid it, so I don't think it's that unreasonable to only support one. Mostly I like that supporting *only* `class << self` removes the name collisions issue (item 2). Definitely interested in @jemmaissroff's thoughts here!

> 2. In the current code, I think you are correct that only the `self.foo` can be memoized. Should I add failing test case for this configuration and look for a way to fix it? Or should this be a follow-on PR?

It's probably a good idea to add a test for this so we have the current behavior documented. I don't know if it's something worth fixing though (assuming we support both class method options). Also interested in @jemmaissroff's thoughts.

jemmaissroff:
Do we know a solution path to be able to implement the `self.foo` class method definitions _and_ avoid method name collision for memoizing `self.foo` _and_ `#foo`?

The way I'm understanding this discussion so far is a question of whether we'd rather prioritize both class method definitions and not allow memoization of class and instance methods of the same name, or prioritize allowing memoization of class and instance methods with the same name and not allow the `self.foo` class method definitions. 

I am wondering if we know of a way to allow us both of these and therefore don't need to prioritize between them? Or if we should continue the conversation around their relative priority? Or if I'm misinterpreting this entire discussion....

ms-ati:
As discussed in Slack, we will:

**Separate the "self." class method API from instance method**

```ruby
class Example
  prepend MemoWise

  def self.foo; end
  memo_wise self: :foo
end
```

**Reduce the "magic" and complexity by knowing ahead of time where method should be**

**Extend same API to the other methods**
[ ] `preset_memo_wise`
[ ] `reset_memo_wise`

ms-ati:
Closing this original "spike" PR in favor of #83

PR #74, opened 2020-10-23
Raise error when memoizing methods with block args

### LINKED ISSUE #39: Handle block arguments
- Explicit block, raise an argument error at setup of memoization
- Implicit block, passed in, ignore for the purposes of memoization?
  - Why: We’d have to check `block_given?` at every call, which is an unacceptable performance cost to ensure we are rejecting something we documented we don’t support
- Document in API docs and README that we don’t support memoizing methods which take blocks
  - Could suggest as a workaround defining the method to take a proc arg as a parameter which we can memoize (add a spec here to verify this works as expected)
    - is this necessary?

jemmaissroff:
This PR:
* Raises an `ArgumentError` when memoizing methods which have explicit block parameters
* Adds specs confirming we can't memoize methods with explicit or implicit block params
* Adds spec confirming we _can_ memoize methods with named proc params
* Adds documentation around memoizing methods with block params

Resolves #39.
 
**Before merging:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,11 @@
 # `MemoWise`
 
 [![Tests](https://github.com/panorama-ed/memo_wise/workflows/Main/badge.svg)](https://github.com/panorama-ed/memo_wise/actions?query=workflow%3AMain)
-[![Code Coverage](https://codecov.io/gh/panorama-ed/memo_wise/branch/main/graph/badge.svg)](https://codecov.io/gh/panorama-ed/memo_wise)
+[![Code Coverage](https://codecov.io/gh/panorama-ed/memo_wise/branch/main/graph/badge.svg)](https://codecov.io/gh/panorama-ed/memo_wise/branches/main)
 [![Yard Docs](http://img.shields.io/badge/yard-docs-blue.svg)](http://rubydoc.info/github/panorama-ed/memo_wise)
 [![Inline docs](http://inch-ci.org/github/panorama-ed/memo_wise.svg?branch=main)](http://inch-ci.org/github/panorama-ed/memo_wise)
 [![Gem Version](https://img.shields.io/gem/v/memo_wise.svg)](https://rubygems.org/gems/memo_wise)
 [![Gem Downloads](https://img.shields.io/gem/dt/memo_wise.svg)](https://rubygems.org/gems/memo_wise)
-[![Code Coverage](https://codecov.io/gh/panorama-ed/memo_wise/branch/main/graph/badge.svg)](https://codecov.io/gh/panorama-ed/memo_wise/branches/main)
 
 
 ## Why `MemoWise`?
@@ -71,6 +70,8 @@ ex.slow_value(3) # => 4 # Returns immediately because the result is memoized
 ex.reset_memo_wise # Resets all memoized results for all methods on ex
 ```
 
+Methods which take implicit or explicit block arguments cannot be memoized.
+
 For more usage details, see our detailed [documentation](#documentation).
 
 ## Benchmarks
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,12 +1,11 @@
 # `MemoWise`
 
 [![Tests](https://github.com/panorama-ed/memo_wise/workflows/Main/badge.svg)](https://github.com/panorama-ed/memo_wise/actions?query=workflow%3AMain)
-[![Code Coverage](https://codecov.io/gh/panorama-ed/memo_wise/branch/main/graph/badge.svg)](https://codecov.io/gh/panorama-ed/memo_wise)
+[![Code Coverage](https://codecov.io/gh/panorama-ed/memo_wise/branch/main/graph/badge.svg)](https://codecov.io/gh/panorama-ed/memo_wise/branches/main)
```
jemmaissroff, line 4: I think this was a rebase issue, but we have the codecov badge twice now, so removing the second occurrence

lib/memo_wise.rb
```diff
@@ -114,6 +114,38 @@ def self.has_kwarg?(method) # rubocop:disable Naming/PredicateName
...
+  #
+  #     def block_args(&block)
+  #       block.call
+  #     end
+  #   end
+  #
+  #   MemoWise.has_block_arg?(Example.instance_method(:position_args)) #=> false
+  #
+  #   MemoWise.has_block_arg?(Example.instance_method(:block_args)) #=> true
+  #
```
jemmaissroff, line 144: These docs are maybe overkill, just copying what we had for `MemoWise.has_args?` and `MemoWise.has_kwargs?``

lib/memo_wise.rb
```diff
@@ -208,6 +245,9 @@ def #{method_name}#{args_str}
   #            calls to the same method with the same set of arguments will then
   #            return that saved result.
   #
+  #   Methods which implicitly or explicitly take block arguments cannot be
+  #   memoized.
```
jemmaissroff, line 249: Let me know if you think I should add documentation around the proc workaround. I _think_ this is an unlikely case, and not worth documenting, but not a strong opinion.

README.md
JacobEvelyn: Ah whoops! I think this was me, sorry!

lib/memo_wise.rb
```diff
@@ -114,6 +114,38 @@ def self.has_kwarg?(method) # rubocop:disable Naming/PredicateName
...
+  #       block.call
+  #     end
+  #   end
+  #
+  #   MemoWise.has_block_arg?(Example.instance_method(:position_args)) #=> false
+  #
+  #   MemoWise.has_block_arg?(Example.instance_method(:block_args)) #=> true
+  #
+  def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
+    method.parameters.last[0] == :block unless method.parameters.empty?
```
JacobEvelyn, line 146: This line can be simplified to any of the following:

```ruby
method(:foo).parameters.last&.first == :block
method(:foo).parameters.last&.[]0 == :block
method(:foo).parameters[-1]&.[]0 == :block
```

lib/memo_wise.rb
JacobEvelyn: I agree; this is not worth documenting.

spec/memo_wise_spec.rb
```diff
@@ -133,6 +135,19 @@ def public_memowise_method
       public :public_memowise_method
 
       def unmemoized_method; end
+
+      def proc_method(proc)
```
JacobEvelyn, line 139: I don't think I see what this adds—to me this method is just the same as any other positional argument method. Thoughts?

jemmaissroff: Yeah, was adding it mostly as documentation because I was unsure how explicit we want to be about the proc workaround. But can remove, totally agreed it's not actually testing anything new!

JacobEvelyn (approved):

PR #83, opened 2020-11-19
Add support for Class method memoization

### LINKED ISSUE #35: Support class method memoization
- [ ] Should also support `reset_memo_wise` and `preset_memo_wise` on class methods

*Note: one approach that might solve this and improve performance overall is using a different instance variable for each method instead of one big hash variable that we instantiate in `initialize`*

jemmaissroff:
Included in this PR is support for memoization of class methods,
which in Ruby are technically instance methods of the singleton
class.

Support class methods defined in two ways.

(1) The "self." syntax:

    class SelfDot
      prepend MemoWise

      def self.foo; end
      memo_wise self: :foo
    end

(2) The "class << self" syntax:

    class ClassShiftSelf
      class << self
        prepend MemoWise

        def foo; end
        memo_wise :foo
      end
    end

Resolves #35 

**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -40,7 +40,7 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
   # [gem](https://rubygems.org/gems/values).
   #
   def initialize(*)
-    @_memo_wise = Hash.new { |h, k| h[k] = {} }
+    MemoWise.create_memo_wise_state!(self)
     super
   end
 
@@ -148,6 +148,80 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
 
   # @private
   #
+  # Returns visibility of an instance method defined on a class.
+  #
+  # @param klass [Class]
+  #   Class in which to find the visibility of an existing *instance* method.
+  #
+  # @param method_name [Symbol]
+  #   Name of existing *instance* method find the visibility of.
+  #
+  # @return [:private, :protected, :public]
+  #   Visibility of existing instance method of the class.
+  #
+  # @raise ArgumentError
+  #   Raises `ArgumentError` unless `method_name` is a `Symbol` corresponding
+  #   to an existing **instance** method defined on `klass`.
+  #
+  def self.method_visibility(klass, method_name)
+    if klass.private_method_defined?(method_name)
+      :private
+    elsif klass.protected_method_defined?(method_name)
+      :protected
+    elsif klass.public_method_defined?(method_name)
+      :public
+    else
+      raise ArgumentError, "#{method_name.inspect} must be a method on #{klass}"
+    end
+  end
+
+  # @private
+  #
+  # Find the original class for which the given class is the corresponding
+  # "singleton class".
+  #
+  # See https://stackoverflow.com/questions/54531270/retrieve-a-ruby-object-from-its-singleton-class
+  #
+  # @param klass [Class]
+  #   Singleton class to find the original class of
+  #
+  # @return Class
+  #   Original class for which `klass` is the singleton class.
+  #
+  # @raise ArgumentError
+  #   Raises if `klass` is not a singleton class.
+  #
+  def self.original_class_from_singleton(klass)
+    unless klass.singleton_class?
+      raise ArgumentError, "Must be a singleton class: #{klass.inspect}"
+    end
+
+    # Search ObjectSpace
+    #   * 1:1 relationship of singleton class to original class is documented
+    #   * Performance concern: searches all Class objects
+    #     But, only runs at load time
+    ObjectSpace.each_object(Class).find { |cls| cls.singleton_class == klass }
+  end
+
+  # @private
+  #
+  # Create initial mutable state to store memoized values if it doesn't
+  # already exist
+  #
+  # @param [Object] obj
+  #   Object in which to create mutable state to store future memoized values
+  #
+  def self.create_memo_wise_state!(obj)
+    unless obj.instance_variables.include?(:@_memo_wise)
+      obj.instance_variable_set(
+        :@_memo_wise,
+        Hash.new { |h, k| h[k] = {} }
+      )
+    end
+  end
+
+  # @private
+  #
   # Private setup method, called automatically by `prepend MemoWise` in a class.
   #
   # @param target [Class]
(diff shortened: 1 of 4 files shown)


lib/memo_wise.rb
```diff
@@ -146,6 +146,95 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+  def self.original_class_from_singleton(klass)
+    unless klass.singleton_class?
+      raise ArgumentError, "Must be a singleton class: #{klass.inspect}"
+    end
+
+    # Option 1: Derive from `#to_s` of the singleton class
+    #   * Works unless original class NOT assigned to constant
+    #   * Depends on an undocumented naming convention in Ruby, where singleton
+    #     class of `Foo` will return "#<Class:Foo>" from `#to_s` method.
+    name = klass.to_s.match(/#<Class:([^>]+)>/)&.[](1)
```
JacobEvelyn, line 203: Comment copied from previous PR:

---

Is there a risk of this finding something but it not being the class we expect? We could mitigate this with:

```
Object.const_get(name).tap do |original|
  raise NameError unless original.singleton_class == klass
end
```

or just remove the `Option 1` approach entirely? Especially if this runs at load time I wonder if it's really worth it to optimize this so much, but I could be convinced either way.

lib/memo_wise.rb
```diff
@@ -146,6 +146,95 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+    raise NameError if name.nil?
+
+    Object.const_get(name) # raises NameError if not found
+  rescue NameError
+    # Option 2: Search ObjectSpace
+    #   * Works in all cases, so makes a good fallback option
+    #   * 1:1 relationship of singleton class to original class is documented
+    #   * Even though this only runs at load time, searches all Class objects
+    ObjectSpace.each_object(Class).find do |cls|
+      !cls.singleton_class? && cls.singleton_class == klass
```
JacobEvelyn, line 213: On the previous PR you (@jemmaissroff) had noted that we could maybe remove the `!cls.singleton_class?` check here.

lib/memo_wise.rb
```diff
@@ -163,25 +252,46 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+        end
+
+        # In Ruby, "class methods" are implemented as normal instance methods
+        # on the "singleton class" of a given Class object, found via
+        # {Class#singleton_class}.
+        # See: https://medium.com/@leo_hetsch/demystifying-singleton-classes-in-ruby-caf3fa4c9d91
+        klass = self
+        singleton = klass.singleton_class
+
+        if singleton.private_instance_methods.include?(method_name) ||
```
JacobEvelyn, line 275: On the previous PR, you (@jemmaissroff) had commented that we might be able to just use

```ruby
if singleton.instance_method(method_name)
```

jemmaissroff: So if it's not an instance method, it'll raise a `NameError`, which means the path I see is to surround this in a `begin / rescue` block. I think that's a little less clean than checking if it's in `private_instance_methods` or `instance_methods`. Also, in digging into this, I realized we didn't have test coverage of private class methods, so I added that

lib/memo_wise.rb
jemmaissroff: I removed option 1. I agree with you here that I'm not sure the optimization is worth the additional complexity

JacobEvelyn (approved):
LGTM! Would prefer to have @ms-ati also review though.

lib/memo_wise.rb
```diff
@@ -163,25 +236,46 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+           singleton.instance_methods.include?(method_name)
+          MemoWise.create_memo_wise_state!(self)
+          klass = singleton
+        elsif klass.singleton_class?
+          MemoWise.create_memo_wise_state!(
+            MemoWise.original_class_from_singleton(klass)
+          )
+        end
```
ms-ati, line 267: @jemmaissroff Can we re-work and clarify the branching paths of logic so that we first take into account whether the `self:` keyword arg was used? I don't think we should be following any fallback logic anymore?

ms-ati: In other words, in this code, are we still looking for a singleton method first even if `self` was not passed?

lib/memo_wise.rb
ms-ati: I think we need test coverage that we can have the same `method_name` at both class and instance levels, and that here we should be using the presence of the `self:` keyword arg to choose which one to memoize?

lib/memo_wise.rb
```diff
@@ -163,25 +236,46 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+
+          method_name = method_name_or_hash[:self]
+        end
+
+        # In Ruby, "class methods" are implemented as normal instance methods
+        # on the "singleton class" of a given Class object, found via
+        # {Class#singleton_class}.
+        # See: https://medium.com/@leo_hetsch/demystifying-singleton-classes-in-ruby-caf3fa4c9d91
+        klass = self
+        singleton = klass.singleton_class
```
ms-ati, line 257: Should the above entire block be inside a branch where the caller past `self:`?

lib/memo_wise.rb
```diff
@@ -146,6 +146,79 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+  #   Raises if `klass` is not a singleton class.
+  #
+  def self.original_class_from_singleton(klass)
+    unless klass.singleton_class?
+      raise ArgumentError, "Must be a singleton class: #{klass.inspect}"
+    end
+
+    # Search ObjectSpace
+    #   * 1:1 relationship of singleton class to original class is documented
+    #   * Even though this only runs at load time, searches all Class objects
```
ms-ati, line 201:
```suggestion
    #   * Performance concern: searches all Class objects. But, only runs at load time
```

lib/memo_wise.rb
```diff
@@ -163,25 +236,46 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+        end
+
+        # In Ruby, "class methods" are implemented as normal instance methods
+        # on the "singleton class" of a given Class object, found via
+        # {Class#singleton_class}.
+        # See: https://medium.com/@leo_hetsch/demystifying-singleton-classes-in-ruby-caf3fa4c9d91
+        klass = self
+        singleton = klass.singleton_class
+
+        if singleton.private_instance_methods.include?(method_name) ||
```
JacobEvelyn, line 259: I think all we need to change here as per @ms-ati's observation is something like:

```ruby
if method_name_or_hash.is_a?(Hash) && (
  klass.singleton_class.private_instance_methods.include?(method_name) ||
  klass.singleton_class.instance_methods.include?(method_name)
)
```

and then we can delete all/most of the `case` statement above?

jemmaissroff: If we want to keep throwing an error when a key other than `self:` is passed, I think the `case` statement is actually cleaner. But I tried to make the logic in there more separate and explicit, let me know what you think now!

JacobEvelyn (approved):

lib/memo_wise.rb
```diff
@@ -146,6 +146,80 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+  #
+  def self.original_class_from_singleton(klass)
+    unless klass.singleton_class?
+      raise ArgumentError, "Must be a singleton class: #{klass.inspect}"
+    end
+
+    # Search ObjectSpace
+    #   * 1:1 relationship of singleton class to original class is documented
+    #   *  Performance concern: searches all Class objects
+    #      But, only runs at load time
```
ms-ati, line 202:
```suggestion
    #   * Performance concern: searches all Class objects
    #     But, only runs at load time
```

lib/memo_wise.rb
```diff
@@ -163,25 +237,43 @@ def self.has_block_arg?(method) # rubocop:disable Naming/PredicateName
...
+          end
+
+          method_name = method_name_or_hash[:self]
+
+          # In Ruby, "class methods" are implemented as normal instance methods
+          # on the "singleton class" of a given Class object, found via
+          # {Class#singleton_class}.
+          # See: https://medium.com/@leo_hetsch/demystifying-singleton-classes-in-ruby-caf3fa4c9d91
+          MemoWise.create_memo_wise_state!(self)
+          klass = klass.singleton_class
```
ms-ati, line 264: Can we move this line up under the comment? The comment refers to this line `klass.singleton_class`, rather than the create state line?

ms-ati, line 264:
```suggestion
          method_name = method_name_or_hash[:self]
          MemoWise.create_memo_wise_state!(klass)

          # In Ruby, "class methods" are implemented as normal instance methods
          # on the "singleton class" of a given Class object, found via
          # {Class#singleton_class}.
          # See: https://medium.com/@leo_hetsch/demystifying-singleton-classes-in-ruby-caf3fa4c9d91
          klass = klass.singleton_class         
```

spec/memo_wise_spec.rb
```diff
@@ -289,6 +298,162 @@ def implicit_block_method
...
+      context "with an invalid hash key" do
+        let(:class_with_memo) do
+          Class.new do
+            prepend MemoWise
+
+            def self.class_method; end
+          end
+        end
+
+        it "memoizes class methods defined with 'def self.'" do
```
ms-ati, line 390: Wrong description of this test

ms-ati (changes requested):
Looking super good @jemmaissroff ! Just a couple more minor changes requested and then I'll approve! Thanks for doing this work

spec/memo_wise_spec.rb
```diff
@@ -785,4 +959,52 @@ def no_args
...
+      context "when klass is a singleton class of an original class" do
+        let(:klass) { original_class.singleton_class }
+
+        context "when assigned to a constant (normal case)" do
+          let(:original_class) { String }
+
+          it { is_expected.to eq(original_class) }
+        end
+
+        context "when not assigned to a constant (anonymous case)" do
```
ms-ati, line 992: Wait can't we remove and reorganize the tests here, since there is only one case now?

ms-ati (approved):
LGTM pending resolution of one more test question

PR #99, opened 2021-01-04
Support allocated instances

JacobEvelyn:
**Before merging:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,13 +83,13 @@ Memoized value retrieval time using Ruby 2.7.2 and
 
 |Method arguments|**`memo_wise` (0.1.0)**|`memery` (1.3.0)|`memoist` (0.16.2)|`memoized` (1.0.2)|`memoizer` (1.0.3)|
 |--|--|--|--|--|--|
-|`()` (none)|**baseline**|11.62x slower|2.35x slower|1.16x slower|3.01x slower|
-|`(a, b)`|**baseline**|2.00x slower|2.26x slower|1.85x slower|2.01x slower|
-|`(a:, b:)`|**baseline**|2.27x slower|2.45x slower|2.16x slower|2.27x slower|
-|`(a, b:)`|**baseline**|1.57x slower|1.72x slower|1.50x slower|1.57x slower|
-|`(a, *args)`|**baseline**|2.02x slower|2.28x slower|1.99x slower|1.99x slower|
-|`(a:, **kwargs)`|**baseline**|1.93x slower|2.07x slower|1.87x slower|1.92x slower|
-|`(a, *args, b:, **kwargs)`|**baseline**|1.94x slower|2.15x slower|1.92x slower|1.93x slower|
+|`()` (none)|**baseline**|12.27x slower|2.57x slower|1.22x slower|3.22x slower|
+|`(a, b)`|**baseline**|2.00x slower|2.21x slower|1.80x slower|2.01x slower|
+|`(a:, b:)`|**baseline**|2.29x slower|2.41x slower|2.16x slower|2.28x slower|
+|`(a, b:)`|**baseline**|1.58x slower|1.71x slower|1.49x slower|1.56x slower|
+|`(a, *args)`|**baseline**|2.03x slower|2.30x slower|1.95x slower|2.01x slower|
+|`(a:, **kwargs)`|**baseline**|1.94x slower|2.08x slower|1.88x slower|1.90x slower|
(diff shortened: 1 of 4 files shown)


lib/memo_wise.rb
```diff
@@ -236,6 +238,20 @@ def self.create_memo_wise_state!(obj)
   #
   def self.prepended(target) # rubocop:disable Metrics/PerceivedComplexity
     class << target
+      # Allocator to setup memoization state before
```
JacobEvelyn, line 241: I really wasn't sure of the best way to document this method. Suggestions welcome.

lib/memo_wise.rb
```diff
@@ -285,12 +301,6 @@ def memo_wise(method_name_or_hash) # rubocop:disable Metrics/PerceivedComplexity
         # hash key is just the method name.
         if method.arity.zero?
           klass.module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
-            # def foo
```
JacobEvelyn, line 288: Not sure why these were here but they looked like the result of testing?

ms-ati: Isn't this is what Rubocop asks for demonstrating what a dynamically-evaluated method will look like?

spec/memo_wise_spec.rb
```diff
@@ -304,6 +293,17 @@ def self.with_positional_args(a, b) # rubocop:disable Naming/MethodParameterName
...
+    context "when instances are created with Class#allocate" do
+      let(:instance) { class_with_memo.allocate }
+
+      it "memoizes correctly" do
+        instance.no_args
+
+        expect(instance.no_args_counter).to eq(1)
+        expect(instance.no_args_counter).to eq(0)
+      end
+    end
```
ms-ati, line 305: Did this test correctly fail without the change?

lib/memo_wise.rb
ms-ati: Love it! Can we also add links to:

Documentation of `Class#allocate`
- https://ruby-doc.org/core-2.7.2/Class.html#method-i-allocate

Relevant call to `Class#allocate` in ActiveRecord 5
- https://github.com/rails/rails/blob/a395c3a6af1e079740e7a28994d77c8baadd2a9d/activerecord/lib/active_record/persistence.rb#L411

ms-ati:
Some questions up, but overall awesome find on `Class#allocate`! A corner of Ruby that TIL!

ms-ati (changes requested):
Hi Jake, can we restore the commented versions of the generated code, and add a comment explaining their presence if that's helpful?

lib/memo_wise.rb
JacobEvelyn: Haha yep I just discovered that. 👍

spec/memo_wise_spec.rb
JacobEvelyn: This was a typo. You reviewed too soon! 😬

README.md
```diff
@@ -83,13 +83,13 @@ Memoized value retrieval time using Ruby 2.7.2 and
...
 |Method arguments|**`memo_wise` (0.1.0)**|`memery` (1.3.0)|`memoist` (0.16.2)|`memoized` (1.0.2)|`memoizer` (1.0.3)|
 |--|--|--|--|--|--|
-|`()` (none)|**baseline**|11.62x slower|2.35x slower|1.16x slower|3.01x slower|
-|`(a, b)`|**baseline**|2.00x slower|2.26x slower|1.85x slower|2.01x slower|
-|`(a:, b:)`|**baseline**|2.27x slower|2.45x slower|2.16x slower|2.27x slower|
-|`(a, b:)`|**baseline**|1.57x slower|1.72x slower|1.50x slower|1.57x slower|
-|`(a, *args)`|**baseline**|2.02x slower|2.28x slower|1.99x slower|1.99x slower|
-|`(a:, **kwargs)`|**baseline**|1.93x slower|2.07x slower|1.87x slower|1.92x slower|
-|`(a, *args, b:, **kwargs)`|**baseline**|1.94x slower|2.15x slower|1.92x slower|1.93x slower|
+|`()` (none)|**baseline**|12.27x slower|2.57x slower|1.22x slower|3.22x slower|
```
JacobEvelyn, line 86: Benchmarks from: https://github.com/panorama-ed/memo_wise/pull/99/checks?check_run_id=1647179824

ms-ati (approved):

PR #101, opened 2021-01-05
Add official testing for Ruby 3.0

JacobEvelyn:
**Before merging:**


README.md
```diff
@@ -78,20 +78,23 @@ For more usage details, see our detailed [documentation](#documentation).
 
 ## Benchmarks
```
JacobEvelyn, line 79: I'd recommend viewing either the rich diff or the end result rendered as markdown.

benchmarks/benchmarks.rb
```diff
@@ -115,7 +121,7 @@ def positional_splat_keyword_and_double_splat_args(a, *args, b:, **kwargs)
     # retrieval time.
     instance.no_args
 
-    x.report("#{benchmark_gem.benchmark_name}: no_args") { instance.no_args }
+    x.report("#{benchmark_gem.benchmark_name}: ()") { instance.no_args }
```
JacobEvelyn, line 124: I updated the names of benchmarks to make it clearer where they go in the README.

benchmarks/benchmarks.rb
```diff
@@ -199,9 +205,11 @@ def positional_splat_keyword_and_double_splat_args(a, *args, b:, **kwargs)
     ARGUMENTS.each { |a, b| instance.keyword_and_double_splat_args(a: a, b: b) }
 
     x.report(
-      "#{benchmark_gem.benchmark_name}: keyword_and_double_splat_args"
+      "#{benchmark_gem.benchmark_name}: (a:, **kwargs)"
     ) do
-      ARGUMENTS.each { |a, b| instance.positional_and_splat_args(a: a, b: b) }
```
JacobEvelyn, line 204: Looks like this was a bug!

benchmarks/benchmarks.rb
```diff
@@ -216,15 +224,15 @@ def positional_splat_keyword_and_double_splat_args(a, *args, b:, **kwargs)
...
+      instance.positional_splat_keyword_and_double_splat_args(a, b, b: b, a: a)
     end
 
     x.report(
-      "#{benchmark_gem.benchmark_name}: "\
-        "positional_splat_keyword_and_double_splat_args"
+      "#{benchmark_gem.benchmark_name}: (a, *args, b:, **kwargs)"
     ) do
       ARGUMENTS.each do |a, b|
-        instance.positional_and_splat_args(a, b, a: a, b: b)
```
JacobEvelyn, line 227: Another bug!

benchmarks/benchmarks.rb
```diff
@@ -216,15 +224,15 @@ def positional_splat_keyword_and_double_splat_args(a, *args, b:, **kwargs)
     # Run once with each set of arguments to memoize the result values, so our
     # benchmark only tests memoized retrieval time.
     ARGUMENTS.each do |a, b|
-      instance.positional_splat_keyword_and_double_splat_args(a, b, a: a, b: b)
+      instance.positional_splat_keyword_and_double_splat_args(a, b, b: b, a: a)
```
JacobEvelyn, line 227: I switched the order of these args because `b: b` is the keyword arg and `a: a` is `**kwargs` and I'm worried that `a: a, b: b` might put them both in through `**kwargs` and bypass the keyword arg.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Logo
 - Memoization of class methods
 - Support for instances created with `Class#allocate`
+- Official testing and benchmarks for Ruby 3.0
```
jemmaissroff, line 14: Worth noting that the code itself was also upgraded to have a ruby version of 3.0?

benchmarks/benchmarks.rb
```diff
@@ -2,13 +2,17 @@
...
 require "memo_wise"
-require "memoist"
-require "memoized"
-require "memoized/version" # Memoized::VERSION does not get loaded above.
-require "memoizer"
-require "memoizer/version" # Memoizer::VERSION does not get loaded above.
+
+# Some gems do not yet work in Ruby 3 so we only require them if they're loaded
+# in the Gemfile.
+%w[memery memoist].each { |gem| require gem if Gem.loaded_specs.key?(gem) }
```
jemmaissroff, line 9: No strong preference here either, but just looking again, could we include `memoized` and `memoizer` in the first line to make the second line also a one liner?
```
%w[memery memoist memoized memoizer].each { |gem| require gem if Gem.loaded_specs.key?(gem) }
# The VERSION constant does not get loaded above. 
%w[memoized memoizer].each { |gem| require "#{gem}/version"  if Gem.loaded_specs.key?(gem) }
```

jemmaissroff (changes requested):

CHANGELOG.md
JacobEvelyn: Do you mean the `.ruby-version` file? I'm not sure that's useful, since that's only used for local development and not users of the gem or our automated tests. Or am I misunderstanding?

benchmarks/benchmarks.rb
JacobEvelyn: Nice suggestion! Unfortunately those lines are too long. The best we could do would be:

```ruby
# Some gems do not yet work in Ruby 3 so we only require them if they're loaded
# in the Gemfile.
%w[memery memoist memoized memoizer].

# The VERSION constant does not get loaded above.
%w[memoized memoizer].
  each { |gem| require "#{gem}/version" if Gem.loaded_specs.key?(gem) }
```

Up to you; I'm happy with whichever you prefer.

CHANGELOG.md
jemmaissroff: No you're totally right, good point, it's internal to the gem.

benchmarks/benchmarks.rb
jemmaissroff: I think I like this better than what we have, if you don't mind changing it! Reads cleaner to me

jemmaissroff (changes requested):

ms-ati (approved):
Looks great!

jemmaissroff (approved):

PR #105, opened 2021-01-21
Removes raising error when memoizing methods which take block args

jemmaissroff:
This PR no longer raises an error when methods which take explicit block args are memoized. 

**Before merging:**


JacobEvelyn:
✅ if you update the benchmarks

JacobEvelyn (approved):

PR #111, opened 2021-02-02
Fix reset_memo_wise support for private methods

### LINKED ISSUE #110: Bug: private method support in reset_memo_wise
Via @JacobEvelyn on Slack:

My branch to port Rainbow uncovered another MemoWise bug 🙂  [This line](https://github.com/panorama-ed/memo_wise/blob/main/lib/memo_wise.rb#L515) should use `respond_to?(method_name, true)` instead of `respond_to?(method_name)` because by default it does not include private methods.

I don't have a ton of time this week—would either of you have time to make a fix?

ms-ati:
Fixes #110


JacobEvelyn (approved):
LGTM as long as that test failed beforehand.

PR #114, opened 2021-02-11
Prepare for initial gem release

ms-ati:
Add Rakefile with "bundler/gem_tasks"
Add instructions for releasing to README.md
Add authors and emails to gemspec

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,30 @@ https://github.com/panorama-ed/memo_wise. This project is intended to be a safe,
 welcoming space for collaboration, and contributors are expected to adhere to
 the [code of conduct](https://github.com/panorama-ed/memo_wise/blob/main/CODE_OF_CONDUCT.md).
 
+## Releasing
+
+To make a new release of `MemoWise` to
+[RubyGems](https://rubygems.org/gems/memo_wise), first install the release
+dependencies (e.g. `rake`) as follows:
+
+```shell
+bundle config set --local with 'release'
+bundle install
+```
+
+Then carry out these steps:
+
+1. Update `CHANGELOG.md`:
+   - Add an entry for the upcoming version _x.y.z_
+   - Move content from _Unreleased_ to the upcoming version _x.y.z_
+   - Commit with title `Update CHANGELOG.md for x.y.z`
+
+2. Update `lib/memo_wise/version.rb`
+   - Replace with upcoming version _x.y.z_
+   - Commit with title `Bump version to x.y.z`
+
+3. `bundle exec rake release`
+
 ## License
 
 The gem is available as open source under the terms of the [MIT License](https://opensource.org/licenses/MIT).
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -155,6 +155,29 @@ https://github.com/panorama-ed/memo_wise. This project is intended to be a safe,
 welcoming space for collaboration, and contributors are expected to adhere to
 the [code of conduct](https://github.com/panorama-ed/memo_wise/blob/main/CODE_OF_CONDUCT.md).
 
+## Releasing
+
+To make a new release of `MemoWise` to Rubygems, first install the release
```
jemmaissroff, line 160: Would it make sense to link RubyGems here? Or maybe the memo_wise RubyGems page once it's up?

jemmaissroff (changes requested):

README.md
JacobEvelyn: While we're talking about this line, tiny nitpick: `Rubygems` -> `RubyGems`

README.md
```diff
@@ -155,6 +155,29 @@ https://github.com/panorama-ed/memo_wise. This project is intended to be a safe,
...
+dependencies:
+
+```shell
+bundle config set --local with 'release'
```
JacobEvelyn, line 164: I'm not sure I understand—why is this necessary?

JacobEvelyn: Ohh, this is so that `rake release` works?

README.md
```diff
@@ -155,6 +155,29 @@ https://github.com/panorama-ed/memo_wise. This project is intended to be a safe,
...
+1. Update `CHANGELOG.md`:
+   - Add an entry for the upcoming version _x.y.z_
+   - Move content from _Unreleased_ to the upcoming version _x.y.z_
+   - Commit with title `Update CHANGELOG.md for x.y.z`
+
+2. Update `lib/memo_wise/version.rb`
+   - Replace with upcoming version _x.y.z_
+   - Commit with title `Bump version to x.y.z`
+
+3. `rake release`
```
JacobEvelyn, line 179: Does this need to be `bundle exec rake release`?

README.md
JacobEvelyn, line 164: Do we need to undo this `bundle config set` afterwards?

README.md
ms-ati: Updated to fix capitalization and add link!

README.md
ms-ati: Yes -- I've added some clarification above, is this good now?

README.md
ms-ati: Updated

README.md
ms-ati: I would not think so? This just means that your local `bundle config` now installs the group `release` for this `Gemfile` -- what would be the benefit of undoing that?

jemmaissroff (approved):

README.md
JacobEvelyn: nvm, you're right.

JacobEvelyn (approved):

PR #116, opened 2021-02-11
Bump version to 0.3.0

ms-ati:
I'll also rebase this one after the previous two PRs are merged.

This is the archetypal final PR before running `rake release`, which does nothing other than officially bumping the version number.

In case you are wondering "when do we create a git tag?", actually the rake task will do that for us.


jemmaissroff (approved):
LGTM once the first two merge!

PR #125, opened 2021-04-01
Allow keyword arguments in initializers when including MemoWise

seckenrode:
Currently on ruby 3.0, prepending MemoWise into a class which accepts keyword arguments
in its initializer causes an ArgumentError to be thrown:

ArgumentError (wrong number of arguments (given 1, expected 0; required keyword: :kwarg))

This change allows both positional and keyword arguments to initializers in classes which
use MemoWise.  This is compatible with versions before 3.0 as well.

**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 - Documentation of confusing module test behavior
+- Allow including MemoWise in classes with keyword arguments in the initializer
 
 ## [0.3.0] - 2021-02-11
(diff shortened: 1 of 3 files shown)


seckenrode:
Apparently this needs some extra special handling for ruby < 2.7.  Whee!

ms-ati:
Thank you @seckenrode! This looks awesome. 🥇 

I'm sure I speak for @jemmaissroff and @JacobEvelyn when I say that (a) We're happy to hear from you and say hi!, and (b) THANK YOU for finding, documenting, and fixing this tricky variance between Ruby versions!

LGTM

ms-ati (approved):
LGTM

PR #130, opened 2021-04-26
Refactor specs to share setup among instance + class contexts

### LINKED ISSUE #129: Refactor test setup to share among instance and class contexts

ms-ati:
Closes #129

More steps:
  * Refactor specs: use comments to break up sections of specs
  * Refactor specs: move instance method tests into contexts
  * Refactor specs: extract shared examples for #memo_wise
  * Extract definition of instance methods for testing
  * Extract to a support file: definition of methods for testing
  * Fix common test setup among instance/class/class-scope contexts
  * Run shared examples for class << self context
  * Simplify test method setup by reusing instance for class << self

### DIFF
--- a/spec/memo_wise_spec.rb
+++ b/spec/memo_wise_spec.rb
@@ -3,599 +3,865 @@
 require "values"
 
 RSpec.describe MemoWise do
-  let(:class_with_memo) do
-    Class.new do
-      prepend MemoWise
+  ##
+  # Shared contexts for setting up for testing:
+  #   * Instance methods
+  #   * Class methods defined via: 'def self.foo ...'
+  #   * Class methods defined via: 'class << self; def foo ...'
+  ##
 
-      def no_args_counter
-        @no_args_counter || 0
-      end
+  shared_context "with context for instance methods" do
+    let(:instance) { class_with_memo.new }
 
-      def with_positional_args_counter
-        @with_positional_args_counter || 0
-      end
+    let(:class_with_memo) do
+      Class.new do
+        prepend MemoWise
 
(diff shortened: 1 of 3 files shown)


spec/memo_wise_spec.rb
```diff
@@ -3,1071 +3,917 @@
...
+            protected_memowise_method_counter + 1
+          "protected_memowise_method"
+        end
+        protected :protected_memowise_method
+        memo_wise :protected_memowise_method
 
-      def with_positional_and_keyword_args_counter
-        @with_positional_and_keyword_args_counter || 0
-      end
+        def self.class_positional_args_counter
```
JacobEvelyn, line 37: Comments explaining why these methods are here would be really helpful.

spec/memo_wise_spec.rb
```diff
@@ -3,1071 +3,917 @@
...
-      def true_method_counter
-        @true_method_counter || 0
-      end
+        def instance_with_keyword_args_counter
+          @instance_with_keyword_args_counter || 0
+        end
 
-      def nil_method_counter
-        @nil_method_counter || 0
+        def with_keyword_args(a:, b:) # rubocop:disable Naming/MethodParameterName
```
JacobEvelyn, line 63: Same here.

spec/memo_wise_spec.rb
ms-ati: Comments added!

spec/memo_wise_spec.rb
ms-ati: Comments added!

JacobEvelyn (approved):

PR #134, opened 2021-04-28
Add class method for preset memo wise

### LINKED ISSUE #86: Class method for `preset_memo_wise`

ms-ati:
Steps:
  * Refactor specs to extract shared examples
  * Add class-level specs for .preset_memo_wise
  * Implement class-level .preset_memp_wise

Closes #86

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -402,6 +402,27 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
         klass.send(visibility, method_name)
       end
     end
+
+    unless target.singleton_class?
+      # Create certain class methods to implement .preset_memo_wise
+      %i[
+        preset_memo_wise
+        validate_memo_wised!
+        validate_params!
+        fetch_key
+      ].each do |method_name|
+        # Like calling 'module_function', but original method stays public
+        target.define_singleton_method(
+          method_name,
+          MemoWise.instance_method(method_name)
+        )
+
+        # Make private the class method copies of private instance methods
+        unless MemoWise.public_method_defined?(method_name)
+          target.singleton_class.send(:private, method_name)
+        end
+      end
+    end
   end
 
   ##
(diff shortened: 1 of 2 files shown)


lib/memo_wise.rb
```diff
@@ -351,6 +351,27 @@ def #{method_name}#{args_str}
...
+        fetch_key
+      ].each do |method_name|
+        # Like calling 'module_function', but original method stays public
+        target.define_singleton_method(
+          method_name,
+          MemoWise.instance_method(method_name)
+        )
+
+        # Make private the class method copies of private instance methods
+        unless MemoWise.public_method_defined?(method_name)
```
jemmaissroff, line 370: What's the case where `MemoWise.public_method_defined?(method_name)` would return `true`?

lib/memo_wise.rb
```diff
@@ -559,16 +585,18 @@ def reset_memo_wise(method_name = nil, *args, **kwargs)
 
   # Validates that {.memo_wise} has already been called on `method_name`.
   def validate_memo_wised!(method_name)
+    klass = instance_of?(Class) ? singleton_class : self.class
```
jemmaissroff, line 588: Is it worth extracting this line into a `klass` method if we use it below too?

jemmaissroff (changes requested):

lib/memo_wise.rb
ms-ati: Hi Jemma! In the current code, some of these methods are private helpers and would return false. `preset_memo_wise` is the public method and returns true

lib/memo_wise.rb
ms-ati: Planning to do extraction and refactoring when I repeat this pattern for `reset_memo_wise` if that’s ok?

jemmaissroff (approved):

lib/memo_wise.rb
```diff
@@ -613,16 +639,19 @@ def reset_memo_wise(method_name = nil, *args, **kwargs)
...
       raise ArgumentError, "#{method_name} is not a memo_wised method"
     end
   end
 
   # Returns arguments key to lookup memoized results for given `method_name`.
   def fetch_key(method_name, *args, **kwargs)
-    method = self.class.instance_method(method_name)
+    klass = instance_of?(Class) ? singleton_class : self.class
+    method = klass.instance_method(method_name)
+    has_arg = MemoWise.has_arg?(method)
```
ms-ati, line 654: Oops, this looks like an accident during the rebase, will remove

PR #138, opened 2021-04-30
Fix Marshal serialization

### LINKED ISSUE #137: Classes that use MemoWise can't be serialized with Marshal
```
class Problem
  prepend MemoWise
end

> Marshal.dump(Problem.new)
Traceback (most recent call last):
        2: from (irb):4
        1: from (irb):4:in `dump'
TypeError (can't dump hash with default proc)
```

The issue may be MemoWise's use of a [default proc on the memo hash](https://github.com/panorama-ed/memo_wise/blob/main/lib/memo_wise.rb#L217).

ms-ati:
As pointed out by @honigc in #137, the use of a default proc in the
core MemoWise Hash instance makes serialization using Marshal impossible.

This PR replaces the implementation with a simple Hash, and sets the
default values along the API code paths that previously relied on the
default value.

We cover the case in #reset_memo_wise where specific args are passed,
to avoid regression of a Nil error when attempting to reset args that
are not yet memoized, as this error would become possible now that
there is no default value proc in the Hash.

Closes #137

**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -212,10 +212,7 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
   # @return [Object] the passed-in obj
   def self.create_memo_wise_state!(obj)
     unless obj.instance_variables.include?(:@_memo_wise)
-      obj.instance_variable_set(
-        :@_memo_wise,
-        Hash.new { |h, k| h[k] = {} }
-      )
+      obj.instance_variable_set(:@_memo_wise, {})
     end
 
     obj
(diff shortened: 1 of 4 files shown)


honigc (approved):

lib/memo_wise.rb
```diff
@@ -299,7 +296,7 @@ def memo_wise(method_name_or_hash) # rubocop:disable Metrics/PerceivedComplexity
         if method.arity.zero?
           klass.module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
             # def foo
-            #   @_memo_wise.fetch(:foo}) do
+            #   @_memo_wise.fetch(:foo) do
```
ms-ati, line 299: Fix a pre-existing typo in the comment demonstrating what the resulting code is supposed to look like

lib/memo_wise.rb
```diff
@@ -551,7 +555,7 @@ def reset_memo_wise(method_name = nil, *args, **kwargs)
     if args.empty? && kwargs.empty?
       @_memo_wise.delete(method_name)
     else
-      @_memo_wise[method_name].delete(fetch_key(method_name, *args, **kwargs))
+      @_memo_wise[method_name]&.delete(fetch_key(method_name, *args, **kwargs))
```
ms-ati, line 558: Added a new test to prove this fixes a real case

PR #146, opened 2021-05-17
Add dependabot for Github Actions to dependabot.yml

ms-ati:
This helps us keep our github actions module versions up to
date over time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,16 +1,20 @@
 version: 2
+
 updates:
-- package-ecosystem: bundler
-  directory: "/"
-  schedule:
-    interval: daily
-    time: "02:00"
-    timezone: America/New_York
-  open-pull-requests-limit: 99
-- package-ecosystem: bundler
-  directory: "/benchmarks"
-  schedule:
-    interval: daily
-    time: "02:00"
-    timezone: America/New_York
-  open-pull-requests-limit: 99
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,16 +1,20 @@
 version: 2
+
 updates:
-- package-ecosystem: bundler
-  directory: "/"
-  schedule:
-    interval: daily
-    time: "02:00"
```
jemmaissroff, line 7: Why did we remove the 2am time?

jemmaissroff (changes requested):

.github/dependabot.yml
ms-ati: Thanks @jemmaissroff ! I wanted to remove unnecessarily specific configuration (in the spirit of _convention over configuration_).

In this case, note that if you don't provide the exact time, it [defaults to 5am UTC](https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates#scheduleinterval), which is currently 1am in eastern time. Do you see benefit in preferring a more specific configuration?

jemmaissroff: Makes sense to me! Wasn't defending either side, just wanted to make sure it was intentional and wasn't a necessary configuration detail! And thanks for checking the default

jemmaissroff (approved):

PR #147, opened 2021-05-17
Testing benchmarks for passing through hash on keyreq args

jemmaissroff:
**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -177,6 +177,10 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
     method.parameters.all? { |type, _| type == :req || type == :keyreq } # rubocop:disable Style/MultipleComparison
   end
 
+  def self.has_only_keyword_required_args?(method) # rubocop:disable Naming/PredicateName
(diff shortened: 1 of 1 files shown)


ms-ati:
@jemmaissroff are you able to reproduce on Ruby 3.0.1 the following error:

```
Branch coverage (99.18%) is below the expected minimum coverage (100.00%)
```

lib/memo_wise.rb
```diff
@@ -341,7 +345,10 @@ def #{method_name}
             end
           END_OF_METHOD
         else
-          if MemoWise.has_only_required_args?(method)
+          if MemoWise.has_only_keyword_required_args?(method)
```
JacobEvelyn, line 348: Just a note—I anticipate this would not be more efficient than the single-arg case. Thoughts on only doing this if it's not a single-`:keyreq` case?

jemmaissroff:
Local benchmarking invalidated this approach. Unfortunately, this is actually slower than what we have now. We presume it's because allocating hashes is less efficient internally than allocating arrays.

PR #154, opened 2021-05-21
Refactor specs to extract shared contexts

ms-ati:
Shared setup contexts:
  * instance methods
  * class methods via 'def self.'
  * class methods via 'class << self'


spec/support/shared_context_for_class_methods_via_self_dot.rb
```diff
@@ -0,0 +1,30 @@
...
+      )
+
+      # Counter for calls to instance method '#with_keyword_args', see below.
+      def instance_with_keyword_args_counter
+        @instance_with_keyword_args_counter || 0
+      end
+
+      # See: "doesn't memoize instance methods when passed self: keyword"
+      #
+      # Used by spec below to verify that `memo_wise self: :with_keyword_args`
```
JacobEvelyn, line 20: Nitpick: `spec below` -> `that spec`

JacobEvelyn (approved):
LGTM if you fix that one comment.

spec/support/shared_context_for_class_methods_via_self_dot.rb
ms-ati, line 20:
```suggestion
      # Used by that spec to verify that `memo_wise self: :with_keyword_args`
```

ms-ati: Fixed!

PR #155, opened 2021-05-21
Refactor specs separate methods

ms-ati:


spec/prepending_initializer_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+RSpec.describe "prepending initializer" do # rubocop:disable RSpec/DescribeClass
+  context "when the class's initializer take arguments" do
```
JacobEvelyn, line 4: Nitpick: `take` -> `takes`

spec/prepending_initializer_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+          def initialize(kwarg:); end
+        end
+      end
+
+      it "does not raise an error when initializing the class" do
+        expect { class_with_memo.new(kwarg: :kw) }.to_not raise_error
+      end
+    end
+
+    context "when it take both positional and keyword arguments" do
```
JacobEvelyn, line 33: Nitpick: `take` -> `takes`

spec/prepending_initializer_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+    end
+
+    context "when the method takes positional arguments, keyword arguments, "\
+              "and a block" do
+      let(:class_with_memo) do
+        Class.new do
+          prepend MemoWise
+
+          def initialize(arg, kwarg:, &block); end
```
JacobEvelyn, line 53: I don't know a ton about how Ruby handles block args but putting them in the params does not make them required:

```ruby
def foo(&block) ; end
foo
=> nil
```

and any method can have a block passed to it:

```ruby
def foo ; end
foo { true }
=> nil
```

Given this, would it make sense to add a `yield` to the `initialize` to make it more closely match real-world usage?

spec/private_apis_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+RSpec.describe MemoWise do # rubocop:disable RSpec/FilePath
+  describe "private APIs" do
+    describe ".method_visibility" do
+      subject { described_class.method_visibility(String, method_name) }
+
+      context "when method_name not a method on klass" do
```
JacobEvelyn, line 8: Nitpick: `not` -> `is not`

spec/private_apis_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      context "when klass is a singleton class of an original class" do
+        let(:klass) { original_class.singleton_class }
+
+        context "when assigned to a constant" do
+          let(:original_class) { String }
+
+          it { is_expected.to eq(original_class) }
+        end
+
+        context "when singleton class #to_s convention not followed" do
```
JacobEvelyn, line 33: Nitpick: `not` -> `is not`

spec/support/shared_context_for_class_methods_via_self_dot.rb
```diff
@@ -0,0 +1,30 @@
...
+      )
+
+      # Counter for calls to instance method '#with_keyword_args', see below.
+      def instance_with_keyword_args_counter
+        @instance_with_keyword_args_counter || 0
+      end
+
+      # See: "doesn't memoize instance methods when passed self: keyword"
+      #
+      # Used by spec below to verify that `memo_wise self: :with_keyword_args`
```
JacobEvelyn, line 20: Nitpick: `spec below` -> `that spec`

JacobEvelyn (commented):
Just some small grammatical nitpicks that were mostly not things you introduced.

spec/prepending_initializer_spec.rb
ms-ati, line 4:
```suggestion
  context "when the class's initializer takes arguments" do
```

ms-ati: Fixed

spec/prepending_initializer_spec.rb
ms-ati, line 33:
```suggestion
    context "when it takes both positional and keyword arguments" do
```

ms-ati: Fixed

spec/prepending_initializer_spec.rb
ms-ati: I see your point! In this particular case, I see `block.call` as being more in keeping with the intent, how's that?

ms-ati, line 53:
```suggestion
          def initialize(arg, kwarg:, &block)
            block.call
          end
```

ms-ati: Changed, is this good?

spec/private_apis_spec.rb
ms-ati, line 8:
```suggestion
      context "when method_name is not a method on klass" do
```

ms-ati: Fixed

spec/private_apis_spec.rb
ms-ati, line 33:
```suggestion
        context "when singleton class #to_s convention is not followed" do
```

ms-ati: Fixed

spec/support/shared_context_for_class_methods_via_self_dot.rb
ms-ati, line 20:
```suggestion
      # Used by that spec to verify that `memo_wise self: :with_keyword_args`
```

ms-ati: Fixed

spec/prepending_initializer_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    context "when the method takes positional arguments, keyword arguments, "\
+              "and a block" do
+      let(:class_with_memo) do
+        Class.new do
+          prepend MemoWise
+
+          def initialize(arg, kwarg:, &blk)
+            blk.call(arg, kwarg) # rubocop:disable Performance/RedundantBlockCall
```
JacobEvelyn, line 54: Ah yes you're right this makes more sense.

JacobEvelyn (approved):

PR #163, opened 2021-05-26
Decorate .instance_method to proxy original params

ms-ati:
In Panorama spec helper code 'initialized_double', we observed
that MemoWise interrupts the use of Ruby reflection on method
parameters, by replacing with delegating methods that have overly
generic signatures, like:

    def initialize(...)

    def slow_method(*arg, **kwargs)

After investigation, we had determined that it was not worth
the complexity cost, and potential fragility, to pursue a solution
where we re-create the original method parmeter signatures for our
delegating implementations.

This PR is a minimalistic alternative, which narrowly overrides only
the implementation of {Module#instance_method} so that, in the case
of either #initialize or a method that has had .memo_wise called on it,
we look up the original method and proxy the original result value of
calling UnboundMethod#parameters on it.

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -404,7 +404,7 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
     end
 
     unless target.singleton_class?
-      # Create certain class methods to implement .preset_memo_wise
+      # Create class methods to implement .preset_memo_wise and .reset_memo_wise
       %i[
         preset_memo_wise
         reset_memo_wise
@@ -421,6 +421,42 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
         # Make private the class method copies of private instance methods
         unless MemoWise.public_method_defined?(method_name)
           target.singleton_class.send(:private, method_name)
+        end
+      end
+
+      # Override [Module#instance_method](https://ruby-doc.org/core-3.0.0/Module.html#method-i-instance_method)
+      # to proxy the original `UnboundMethod#parameters` results. We want the
+      # parameters to reflect the original method in order to support callers
+      # who want to use Ruby reflection to process the method parameters,
+      # because our overridden `#initialize` method, and in some cases the
+      # generated memoized methods, will have a generic set of parameters (e.g.
+      # `...` or `*args, **kwargs, &block`), making reflection on method
+      # parameters useless without this.
+      def target.instance_method(symbol)
+        # TODO: Extract this method naming pattern
+        original_memo_wised_name = :"_memo_wise_original_#{symbol}"
+
+        super.tap do |curr_method|
+          # Start with calling the original `instance_method` on `symbol`,
+          # which returns an `UnboundMethod`.
+          #   IF it was replaced by MemoWise,
+          #   THEN find the original method's parameters, and modify current
+          #        `UnboundMethod#parameters` to return them.
+          if symbol == :initialize
+            # For `#initialize` - because `prepend MemoWise` overrides the same
+            # method in the module ancestors, use `UnboundMethod#super_method`
+            # to find the original method.
+            orig_method = curr_method.super_method
+            orig_params = orig_method.parameters
+            curr_method.define_singleton_method(:parameters) { orig_params }
+          elsif private_method_defined?(original_memo_wised_name)
+            # For any memoized method - because the original method was renamed,
+            # call the original `instance_method` again to find the renamed
+            # original method.
+            orig_method = super(original_memo_wised_name)
+            orig_params = orig_method.parameters
+            curr_method.define_singleton_method(:parameters) { orig_params }
+          end
         end
       end
     end
--- a/spec/proxying_original_method_params_spec.rb
+++ b/spec/proxying_original_method_params_spec.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: true
+
+RSpec.describe "proxying original method params" do # rubocop:disable RSpec/DescribeClass
+  describe ".instance_method" do
+    subject { unbound_method&.parameters }
+
+    let(:unbound_method) { class_with_memo.instance_method(method_name) }
+
+    let(:class_with_memo) do
+      Class.new do
+        prepend MemoWise
+
+        def initialize(foo, bar:); end
+
+        DefineMethodsForTestingMemoWise.define_methods_for_testing_memo_wise(
+          target: self,
+          via: :instance
+        )
+
+        def unmemoized_with_positional_and_keyword_args(a, b:) # rubocop:disable Naming/MethodParameterName
+          [a, b]
+        end
+      end
+    end
+
+    context "when #initialize" do
+      let(:method_name) { :initialize }
+      let(:expected_parameters) { [[:req, :foo], [:keyreq, :bar]] } # rubocop:disable Style/SymbolArray
+
+      it "returns expected parameters" do
+        is_expected.to eq(expected_parameters)
+      end
+
+      it "proxies UnboundMethod#parameters via singleton method" do
+        expect(unbound_method.singleton_methods).to eq [:parameters]
+      end
+    end
+
+    context "when #with_optional_positional_and_keyword_args" do
+      let(:method_name) { :with_optional_positional_and_keyword_args }
+      let(:expected_parameters) { [[:opt, :a], [:key, :b]] } # rubocop:disable Style/SymbolArray
+
+      it "returns expected parameters" do
+        is_expected.to eq(expected_parameters)
+      end
+
+      it "proxies UnboundMethod#parameters via singleton method" do
+        expect(unbound_method.singleton_methods).to eq [:parameters]
+      end
+    end
+
+    context "when #unmemoized_method" do
+      let(:method_name) { :unmemoized_method }
+      let(:expected_parameters) { [] }
+
+      it "returns expected parameters" do
+        is_expected.to eq(expected_parameters)
+      end
+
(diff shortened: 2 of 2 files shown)


spec/proxying_original_method_params_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      it "returns parameters proxied from original method" do
+        is_expected.to eq(expected_parameters)
+      end
+    end
+
+    context "when #unmemoized_method" do
+      let(:method_name) { :unmemoized_method }
+      let(:expected_parameters) { [] }
+
+      it "returns parameters proxied from original method" do
```
ms-ati, line 42:
```suggestion
      it "returns parameters directly from the method" do
```

ms-ati:
Re: #165

lib/memo_wise.rb
```diff
@@ -423,6 +423,28 @@ def #{method_name}#{args_str}
           target.singleton_class.send(:private, method_name)
         end
       end
+
+      # Override [Module#instance_method](https://ruby-doc.org/core-3.0.0/Module.html#method-i-instance_method)
+      # to proxy the original `UnboundMethod#parameters` results. This proxying
+      # helps code that uses Ruby reflection over method parameters
+      def target.instance_method(symbol)
```
JacobEvelyn, line 430: Whoa, I wasn't aware you could use `def` like this.

lib/memo_wise.rb
```diff
@@ -423,6 +423,28 @@ def #{method_name}#{args_str}
...
+        original_memo_wised_name = :"_memo_wise_original_#{symbol}"
+        curr_method = super
+
+        orig_method =
+          if symbol == :initialize
+            curr_method.super_method
```
JacobEvelyn, line 436: The fact that this evaluates to `super.super_method` is kind of confusing to me. Would you mind adding some comments here for what this is doing and why we need to special-case it?

lib/memo_wise.rb
```diff
@@ -423,6 +423,28 @@ def #{method_name}#{args_str}
...
+          elsif private_method_defined?(original_memo_wised_name)
+            super(original_memo_wised_name)
+          end
+
+        unless orig_method.nil?
```
JacobEvelyn, line 441: Could you add a comment explaining when `orig_method` will be `nil`? Is it just when there's no `initialize` method defined?

lib/memo_wise.rb
```diff
@@ -423,6 +423,28 @@ def #{method_name}#{args_str}
...
+          orig_params = orig_method.parameters
+          curr_method.define_singleton_method(:parameters) { orig_params }
```
JacobEvelyn, line 443: This approach makes sense to me, but I just want to make sure we're on the same page that we think this is better than just proxying the whole `UnboundMethod` object instead.

JacobEvelyn (changes requested):

lib/memo_wise.rb
ms-ati: I'll add explicit documentation of the three paths:

1.  `initialize` - call `super` which is the original implementation of `Module#instance_method`, get back the `UnboundMethod` which represents our `MemoWise` version of `initialize` -- then ask **it** for `#super_method`, because it know the original method that it is overriding
 
2. Any memo-wised method - we need to look up the original method by it's renamed symbol, and call `super` which is the original implementation of `instance_method`, using that name

3. Anything else - we have `orig_method = nil`

lib/memo_wise.rb
ms-ati: I'll clarify the cases -- just returning `curr_method` when it was not a memo_wised method and not initialize -- can make it clearer

lib/memo_wise.rb
ms-ati: Discussed live - I'm open on this - was leaning towards minimal "lying" to start to avoid unintentionally breaking things

JacobEvelyn (approved):
LGTM if you add more documentation.

spec/proxying_original_method_params_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+
+      it "proxies UnboundMethod#parameters via singleton method" do
+        expect(unbound_method.singleton_methods).to eq [:parameters]
+      end
+    end
+
+    context "when #unmemoized_method" do
+      let(:method_name) { :unmemoized_method }
+      let(:expected_parameters) { [] }
```
jemmaissroff, line 50: Should we have a test case with `unmemoized_methods` and parameters?

lib/memo_wise.rb
```diff
@@ -423,6 +423,28 @@ def #{method_name}#{args_str}
           target.singleton_class.send(:private, method_name)
         end
       end
+
+      # Override [Module#instance_method](https://ruby-doc.org/core-3.0.0/Module.html#method-i-instance_method)
+      # to proxy the original `UnboundMethod#parameters` results. This proxying
+      # helps code that uses Ruby reflection over method parameters
+      def target.instance_method(symbol)
+        original_memo_wised_name = :"_memo_wise_original_#{symbol}"
```
jemmaissroff, line 431: Should we pull this out since we use it elsewhere too? Wouldn't want the case where we change it somewhere and forget to change it here

lib/memo_wise.rb
```diff
@@ -423,6 +423,28 @@ def #{method_name}#{args_str}
...
+        curr_method = super
+
```
jemmaissroff, line 433: I found the below logic a little hard to follow. Wonder what you think of a refactor something like this:

```
        super.tap do |curr_method|
          if symbol == :initialize
            curr_method.define_singleton_method(:parameters) do
              curr_method.super_method.parameters
            end
          elsif private_method_defined?(original_memo_wised_name)
            orig_super = super(original_memo_wised_name)
            curr_method.define_singleton_method(:parameters) do
              orig_super.parameters
            end
          end
        end
```

I think it's cleaner because surrounding the whole thing in a tap makes it explicit that the goal of this method is to amend `super`. It's also more clear to me this way that if the method doesn't fall into either of the cases, we don't amend the method at all.

jemmaissroff (changes requested):

lib/memo_wise.rb
ms-ati: Sure, I'm happy to refactor this way!

lib/memo_wise.rb
ms-ati: Can we delay that until a following PR when I extract all the private API stuff? I don't want to add additional helper methods now, as they pollute the method namespace of the target classes in the current implementation - I'm planning to change that soon -- and had already responded to you on different PRs that we could start extracting more, smaller helper methods at that point?

jemmaissroff: Sounds good - sorry I'm losing track a bit of which PRs are still coming in this series and when we're extracting the helpers. As long as you have a good grasp of when we're doing this, works for me!

ms-ati: Added a TODO for future extraction of the common method naming pattern

lib/memo_wise.rb
ms-ati: Modified per @jemmaissroff above and documented per this discussion, please re-review?

lib/memo_wise.rb
ms-ati: Modifed as discussed, and added comments per @JacobEvelyn below, please re-review?

lib/memo_wise.rb
ms-ati: Please re-review

lib/memo_wise.rb
ms-ati: Agreed to go with just proxying the original parameter values, for minimally-intrusive change

spec/proxying_original_method_params_spec.rb
ms-ati: Added! Please re-review

lib/memo_wise.rb
```diff
@@ -404,7 +404,7 @@ def #{method_name}#{args_str}
     end
 
     unless target.singleton_class?
-      # Create certain class methods to implement .preset_memo_wise
+      # Create class methods to implement .preset_memo_wise and .reset_memo_wise
```
ms-ati, line 407: Unrelated comment clarification

lib/memo_wise.rb
```diff
@@ -425,25 +425,38 @@ def #{method_name}#{args_str}
       end
 
       # Override [Module#instance_method](https://ruby-doc.org/core-3.0.0/Module.html#method-i-instance_method)
-      # to proxy the original `UnboundMethod#parameters` results. This proxying
-      # helps code that uses Ruby reflection over method parameters
+      # to proxy the original `UnboundMethod#parameters` results. We want the
+      # parameters to reflect the original method in order to support callers
+      # who want to use Ruby reflection to process the method parameters,
+      # because the generated memoized method would have a generic set of
```
JacobEvelyn, line 431: Would you mind tweaking this to say something like `... because our overridden `initialize` method, in some cases the generated memoized methods, would have ...`? (Not sure if you're aware but avoiding generic params like `*args` in many cases is how @jemmaissroff and I got some big speedups recently.)

jemmaissroff (approved):

lib/memo_wise.rb
jemmaissroff: Looks great, thanks!

lib/memo_wise.rb
ms-ati: Changed as requested!

Issue #160, opened 2021-05-24
Refactor for understandability


ms-ati:
Before reaching v1.0, we'd like to move from the original code layout of "one huge `memo_wise.rb` for the code, and another huge `memo_wise_spec.rb` for the tests", and towards a model of extracted code and specs that are individually more approachable and readable.

PR #170, opened 2021-06-15
Extract private apis

ms-ati:
In order to increase the clarity and maintainability of the codebase, extract
internal private apis from the main MemoWise module. This should also make
it easier to refactor private api code.

**Before merging:**

~[ ] If this change merits an update to `CHANGELOG.md`, add an entry following Keep a Changelog [guidelines](https://keepachangelog.com/en/1.0.0/) with [semantic versioning](https://semver.org/)~


ms-ati:
Re: #160

lib/memo_wise/internal_api.rb
```diff
@@ -0,0 +1,244 @@
...
+    def validate_memo_wised!(method_name)
+      klass = target.instance_of?(Class) ? target.singleton_class : target.class
+      original_memo_wised_name = :"_memo_wise_original_#{method_name}"
+
+      unless klass.private_method_defined?(original_memo_wised_name)
+        raise ArgumentError, "#{method_name} is not a memo_wised method"
+      end
+    end
+
+    # TODO: Parameter validation for presetting values
```
jemmaissroff, line 236: I wonder if we should just remove this placeholder for now? As far as I know, we won't do this for V1, so probably okay to just add it back when we need it?

jemmaissroff (approved):

lib/memo_wise/internal_api.rb
ms-ati: Ok, I'll just go ahead and remove this now then

ms-ati: Ok, despite what I wrote in slack, I'll go ahead and just remove this now

PR #181, opened 2021-07-08
Update documentation to Ruby 3.0.2 and 2.7.4

JacobEvelyn:
This commit also removes `benchmarks/Gemfile.lock` from
version control, because we explicitly pin all versions
in `benchmarks/Gemfile` anyway, and `benchmarks/Gemfile.lock`
will confusingly only show the gems installed for that
particular Ruby version.

**Before merging:**

- [ ] ~If this change merits an update to `CHANGELOG.md`, add an entry following Keep a Changelog [guidelines](https://keepachangelog.com/en/1.0.0/) with [semantic versioning](https://semver.org/)~ (I don't think this is necessary—thoughts?)


.gitignore
```diff
@@ -1,5 +1,6 @@
 /.bundle/
 /benchmarks/.bundle/
+/benchmarks/Gemfile.lock
```
JacobEvelyn, line 3: To be honest I thought we had already done this.

@jemmaissroff @ms-ati am I correct that we should not do this for the root `Gemfile.lock`, so Dependabot can still use those versions? Should we change the root `Gemfile` to pin everything explicitly instead?

benchmarks/.ruby-version
```diff
@@ -1 +1 @@
-3.0.0
```
JacobEvelyn, line 1: Should we remove this file? Or have a similar file in the root directory?

.gitignore
jemmaissroff: You know much more about Dependabot than me, but it's my understanding that Dependabot needs the root `Gemfile.lock`, so it would make sense to pin versions right?

benchmarks/.ruby-version
jemmaissroff: What are implications of removing it?

jemmaissroff (changes requested):

benchmarks/.ruby-version
jemmaissroff: Discussed in person - let's remove this file since the gem supports multiple ruby versions

jemmaissroff (approved):

PR #187, opened 2021-07-18
Add dry-core Memoizable to the benchmark

Sorc96:
Added the Memoizable module from [dry-core](https://github.com/dry-rb/dry-core) to the benchmark. The 0 and 1 argument benchmarks favour `memo_wise`, while `dry-core` is faster in the others.

I'm not exactly sure what the differences in implementation are. Maybe there is more optimization potential here, or maybe it is simply a matter of prioritizing a certain type of methods.

The table in README.md seems to assume that other gems always run slower, therefore I wasn't sure how to update it. Should the table be restructured somehow or should I calculate the numbers manually? (memo_wise 2x slower => dry-core 0.5x slower)

### DIFF
--- a/benchmarks/Gemfile
+++ b/benchmarks/Gemfile
@@ -7,6 +7,7 @@ ruby ">= 2.7.2"
 gem "benchmark-ips", "2.9.1"
 
 if RUBY_VERSION > "3"
+  gem "dry-core", "0.7.1"
   gem "memery", "1.4.0"
 else
   gem "ddmemoize", "1.0.0"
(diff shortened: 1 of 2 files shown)


jemmaissroff (approved):
Thanks for sharing `dry-core` with us! This PR looks great.

We took a look at the `dry-core` codebase, and found some ideas which we'll incorporate into this gem. We're happy to start making PRs to `dry-core` to incorporate some of our ideas there if it's helpful too.

The benchmarks in the readme are a little wonky to work with, so we'll take care of adding this to the readme and the changelog.

PR #189, opened 2021-07-20
Add further optimizations

JacobEvelyn:
**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "set"
+
 require "memo_wise/internal_api"
 require "memo_wise/version"
 
@@ -154,14 +156,11 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
         # hash key is just the method name.
         if method.arity.zero?
           klass.module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
-            # def foo
-            #   @_memo_wise.fetch(:foo) do
-            #     @_memo_wise[:foo] = _memo_wise_original_foo
-            #   end
-            # end
-
             def #{method_name}
-              @_memo_wise.fetch(:#{method_name}) do
+              output = @_memo_wise[:#{method_name}]
+              if output || @_memo_wise.key?(:#{method_name})
+                output
+              else
                 @_memo_wise[:#{method_name}] = #{original_memo_wised_name}
               end
             end
@@ -176,12 +175,14 @@ module MemoWise # rubocop:disable Metrics/ModuleLength
               type == :req ? name : "#{name}: #{name}"
             end.join(", ")
             call_str = "(#{call_str})"
-            fetch_key = method.parameters.map(&:last)
-            fetch_key = if fetch_key.size > 1
-                          "[#{fetch_key.join(', ')}].freeze"
-                        else
-                          fetch_key.first.to_s
-                        end
+            fetch_key_params = method.parameters.map(&:last)
+            if fetch_key_params.size > 1
+              fetch_key_init =
+                "[:#{method_name}, #{fetch_key_params.join(', ')}].hash"
+              use_hashed_key = true
+            else
+              fetch_key = fetch_key_params.first.to_s
+            end
           else
             # If our method has arguments, we need to separate out our handling
             # of normal args vs. keyword args due to the changes in Ruby 3.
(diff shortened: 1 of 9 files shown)


lib/memo_wise.rb
```diff
@@ -178,7 +184,9 @@ def #{method_name}
             call_str = "(#{call_str})"
             fetch_key = method.parameters.map(&:last)
             fetch_key = if fetch_key.size > 1
-                          "[#{fetch_key.join(', ')}].freeze"
+                          fetch_key_init =
```
jemmaissroff, line 187: I think setting the `fetch_key_init` within the `if/else` block that sets `fetch_key` is a little confusing. What would you think of:

```ruby
if fetch_key.size > 1
  fetch_key_init = "key = [#{fetch_key.join(', ')}].hash"
  fetch_key = "key"
else
  fetch_key = fetch_key.first.to_s
end
```

jemmaissroff (changes requested):
One small comment. Also - is `#hash` faster than `#freeze`?

JacobEvelyn:
> One small comment. Also - is `#hash` faster than `#freeze`?

Yes, without `#hash` we're much slower than `dry-core` (with or without `#freeze`) in many benchmarks. I don't fully understand why it's so much slower but I suspect that when you use a non-primitive object as a key to a hash, Ruby calls `#hash` under the hood, so this just eliminates some indirection (as well as eliminating the work that `#freeze` does).

lib/memo_wise.rb
JacobEvelyn: Yep totally agree this is confusing. I was trying to just get something working before cleaning up the code as I still don't love where the code is even with your change (there's a lot of duplication around `"key"`, for example).

Do you have a sec to pair on this?

lib/memo_wise.rb
```diff
@@ -192,33 +200,36 @@ def #{method_name}
...
-          # because Ruby always copies argument arrays when splatted.
           klass.module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
             # def foo(*args, **kwargs)
-            #   hash = @_memo_wise.fetch(:foo) do
-            #     @_memo_wise[:foo] = {}
-            #   end
-            #   hash.fetch([args, kwargs].freeze) do
-            #     hash[[args, kwargs].freeze] = _memo_wise_original_foo(*args, **kwargs)
+            #   hash = @_memo_wise[:foo] ||= {}
+            #   key = [args, kwargs].hash
```
ms-ati, line 217: To be clear: this does the following:

1. Create the `Array [args, kwargs]`
2. Calls method `Array#hash` on it, which delegates to both args.hash and kwargs.hash internally
3. Throws away the array from (1) as soon as GC runs, no reference is held to it
4. Uses the hash value as the key

IF this is accruate: is there any advantage to trying to avoid the allocation of the 2-element array in 1?

JacobEvelyn: I had the same thought process @ms-ati. I don't know enough about hash functions to feel that confident implementing our own here (i.e. something like `args.hash ^ kwargs.hash`), though I'm open to it. I did look at the [`Array#hash` source](https://ruby-doc.org/core-3.0.1/Array.html#method-i-hash) but was wary of trying to implement something like that in Ruby and missing out on future improvements to the built-in `Array#hash`. Thoughts?

JacobEvelyn: Update: using XOR is indeed a noticeable speed boost. @ms-ati @jemmaissroff do we think this is safe to use or should we worry about hash collisions?

JacobEvelyn: Update: I no longer think we should use XOR, because I just realized it does not handle ordering, so `foo(1, 2)` and `foo(2, 1)` would collide. Still open to other options that avoid the array allocation though!

JacobEvelyn: Based on https://crypto.stackexchange.com/a/55219, @jemmaissroff and I played around with doing things like

```ruby
(args.hash << 64) & kwargs.hash).hash
```

which is indeed faster.

I don't know enough about:
- **hashing** to say whether this is safe from a collision perspective
- **ruby** to say whether 64 is something platform-dependent here

Sorc96: That is definitely not right. It should be `|` instead of `&`. The idea seems to be that we assume the hash function returns a 64 bit number. Shift it by 64 bits to the left to make room for the second number. Then we need to use bitwise `or`, not `and` to put the two numbers together. Then get the has of the result to get back into the range of 64 bits.

I'm certainly not an expert on this, so I can't say if it's a legitimate algorithm or not. But I do know about the [Cantor pairing function](https://en.wikipedia.org/wiki/Pairing_function). It might be worth checking out.

JacobEvelyn: 🤦‍♀️ yes I meant to write `|`—thank you for the correction! I'll look into the Cantor pairing function—thanks for the link! @ms-ati I'm curious if you have any knowledge here.

ms-ati: Hi @JacobEvelyn -- Hmm stepping back, I know the goal is performance to avoid the ephemeral allocation of the `Array`. But using only a `#hash` as the memoization key does invite collision, and investigating the collisions.

Are there any alternatives that **also avoid the allocation** but retain the full ability to avoid collision of memoized inputs?

JacobEvelyn: Update: I don't think we should use

```ruby
(args.hash << 64) | kwargs.hash).hash
```

as I'm seeing lots of hash collisions with that approach.

JacobEvelyn: Going to table this for now and potentially explore it again in #190

jemmaissroff (changes requested):

lib/memo_wise.rb
```diff
@@ -526,9 +553,18 @@ def reset_memo_wise(method_name = nil, *args, **kwargs)
 
     if args.empty? && kwargs.empty?
       @_memo_wise.delete(method_name)
+      @_memo_wise_hashes[method_name]&.each do |hash|
```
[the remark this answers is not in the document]
JacobEvelyn: I think the piece you're missing is that we need to remove all of the memoized values tied to these hashes, and we do that by deleting each one from `@_memo_wise`.

jemmaissroff (approved):
Looks great!

PR #196, opened 2021-07-29
Update CHANGELOG.md for V1.1.0

jemmaissroff:
**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## Unreleased
+### Updated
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -16,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fix usage on module singleton classes
 - Fix usage on module which would be extended by other classes
+### Breaking Changes
```
JacobEvelyn, line 27: Should we add these sections for past sections too?

JacobEvelyn (approved):

CHANGELOG.md
jemmaissroff: I thought about it. I decided not to because I figured that we only publicized the gem after the 1.0.0 release, so it was unlikely people were upgrading from versions before that.

JacobEvelyn: Fine by me. I don't think I see a reason not to add them but I don't feel strongly; either way works.

PR #200, opened 2021-08-05
Sort benchmarks headers to match cells

JacobEvelyn:
**Before merging:**

### DIFF
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -243,6 +243,9 @@ end.each_with_index do |benchmark_json, i|
   memo_wise = benchmark_json.find { _1["name"].include?("MemoWise") }
   benchmark_json.delete(memo_wise)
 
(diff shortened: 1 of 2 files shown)


benchmarks/benchmarks.rb
```diff
@@ -245,7 +245,9 @@ def positional_splat_keyword_and_double_splat_args(a, *args, b:, **kwargs)
 
   # Print headers based on the first benchmark_json
   if i.zero?
-    benchmark_headers = benchmark_json.map do |benchmark_gem|
+    benchmark_headers = benchmark_json.
```
jemmaissroff, line 248: We repeat this `sort_by` code later on line 262. Wondering if it'd make sense to put it into a `sorted_benchmark_json` variable before this `if` so we can use it later too?

jemmaissroff (changes requested):

jemmaissroff (approved):

PR #206, opened 2021-08-11
Speed up tests by reusing temporary classes/modules

### LINKED ISSUE #202: Investigate whether we can speed up tests
Many of our tests run quite quickly (as they should, since they simply execute pure Ruby), but many are quite slow (as in: >1s per test). The slow ones seem to be within `when defined with scope 'class << self'` and `when defined with scope 'module << self'` context blocks, though there may be others I've missed.

Test time currently is a big nuisance when working on features/and speeding up tests would be a big quality-of-life improvement.

JacobEvelyn:
Fixes #202

NOTE: I'm not sure if this is a good idea but it does make our tests much faster.

**Before merging:**

### DIFF
--- a/spec/support/shared_context_for_class_methods_via_class_scope.rb
+++ b/spec/support/shared_context_for_class_methods_via_class_scope.rb
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
 RSpec.shared_context "with context for class methods via scope 'class << self'" do # rubocop:disable Layout/LineLength
-  let(:class_with_memo) do
-    Class.new do
+  # NOTE: This use of `before(:all)` is a performance optimization that shaves
+  # minutes off of our test suite, especially in older versions of Ruby.
+  before(:all) do
+    @_class_with_memo = Class.new do
       class << self
         prepend MemoWise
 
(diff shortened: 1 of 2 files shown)


spec/support/shared_context_for_module_methods_via_class_scope.rb
```diff
@@ -13,4 +15,6 @@ class << self
       end
     end
   end
+
+  let(:module_with_memo) { @_module_with_memo } # rubocop:disable RSpec/InstanceVariable
```
ms-ati, line 19: Does `module_with_memo` need the same Module-level resetting of state that the `class_with_memo` needed?

spec/support/shared_context_for_class_methods_via_class_scope.rb
```diff
@@ -13,4 +15,18 @@ class << self
...
     end
   end
+
+  let(:class_with_memo) do
+    # Because we now have shared state between tests, we need to ensure that we
+    # reset memo_wise, as well as any test state, for each individual test.
+    # rubocop:disable RSpec/InstanceVariable
+    @_class_with_memo.reset_memo_wise
+    @_class_with_memo.instance_variables.each do |var|
+      if @_class_with_memo.instance_variable_get(var).is_a?(Integer)
```
ms-ati, line 25:
```suggestion
      # reset test method counters from DefineMethodsForTestingMemoWise
      if @_class_with_memo.instance_variable_get(var).is_a?(Integer)
```

ms-ati (approved):
LGTM, I left one commit suggestion with a comment line you might want to consider adding to make the "reset counters to 0" test code more understandable

spec/support/shared_context_for_module_methods_via_class_scope.rb
JacobEvelyn: I haven't looked into why, but it does not.

jemmaissroff (approved):

PR #211, opened 2021-09-22
Optimize zero-argument methods

JacobEvelyn:
This commit adds a performance optimization to zero-argument
methods; these methods now have their memoized results stored
in an array rather than a hash. This is the same as the recent
optimization we made for one-argument methods, except because
we store zero-argument results directly (rather than a hash
containing results as with one-argument methods), we need a
sentinel system to differentiate between an unmemoized method
(in which case the lookup will return `nil`) and a memoized
result that also happens to be `nil`.

This optimization offers a slight speedup because array
accesses are faster than hash lookups. In addition, this
optimization is faster even for the "slow path" case in which
the memoized value is falsey, because previously that check
was also performing a hash lookup and our new sentinel system
uses an array for this as well.

**Before merging:**


CHANGELOG.md
```diff
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Updated
-- Improve performance of single-argument methods by using an outer Array instead
-  of a Hash
+- Improve performance of both zero-argument and single-argument methods by using
+  an outer Array instead of a Hash
```
ms-ati, line 12: Nit: I'd split these into two CHANGELOG entries personally, to make it easier to connect to separate changes in the history, but weakly held

ms-ati (approved):
LGTM

PR #213, opened 2021-10-11
[Draft] Trying out a different API

paracycle:
Prepending a module as a public API is quite uncommon. Since MemoWise primarily adds a class method named `memo_wise`, doing an extend is a much more natural API.

Moreover, MemoWise currently does an alias method chain dance when redefining memoized modules which is brittle. A cleaner way of doing that is by building memoization wrappers in a prepended class and calling `super` to delegate to the actual method.

**Before merging:**


ms-ati:
Hi again @paracycle, this PR is growing stale in its current form. If it's not too much trouble, please respond to communicate whether you intend to continue work on this PR, such as to resolve existing conflicts, so that we can discuss the proposed changes in detail?

paracycle:
Apologies that this PR has gone stale (actually, it was not too fresh when I opened it either). I do intend to bring this inline with the latest changes and have a more reviewable and contained PR.

In the meanwhile, seen as this PR was mostly for me to get this alternative approach some visibility, I will close it and I can create a new PR later.

jemmaissroff:
@paracycle I would be happy to pair on getting this PR back up to date if you'd like!

paracycle:
> @paracycle I would be happy to pair on getting this PR back up to date if you'd like!

Yeah, I was going to propose the same. 👍 I'll book some time for us to pair on this.

PR #216, opened 2021-10-27
Remove usages of #hash to avoid potential hash collisions

jemmaissroff:
This PR removes usage of #hash because there is a potential for hash collisions leading to incorrect memoized results in our current implementation. 

See [this issue](https://github.com/dry-rb/dry-core/issues/63) for further explanation.

Still todo on this PR:

- [ ] Once benchmarks run, add content around what we do differently than dry-rb
- [X] If this change merits an update to `CHANGELOG.md`, add an entry following Keep a Changelog [guidelines](https://keepachangelog.com/en/1.0.0/) with [semantic versioning](https://semver.org/)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve performance of single-argument methods by using an outer Array instead
   of a Hash
 ### Fixed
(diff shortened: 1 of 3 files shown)


JacobEvelyn (changes requested):
We no longer need `@_memo_wise_hashes`, right?

jemmaissroff:
Resolved in #219

PR #224, opened 2021-11-10
Fix thread-safety for concurrent uncached calls

jeremyevans:
This updates the sentinel after updating the cache, and check the sentinel before checking the cache.  Otherwise, if there are two concurrent calls for an uncached value in separate threads, a thread-safety issue is possible:

1. Thread 1 sets sentinel value to true
2. Thread 2 gets cached value (nil is not set)
3. Thread 2 checks sentinel value (is true)
4. Thread 2 returns nil instead of correct value
5. Thread 1 sets cached value

This changes the behavior so that in the case of concurrent calls for an uncached value, the original method is called twice instead of in invalid value being returned.

This change may negatively affect performance for cached calls, I haven't benchmarked or tested this.

**Before merging:**

I'll leave the before merging changes to the maintainers if they choose to accept this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-- Fix theoretical thread-safety issue for concurrent calls to an uncached
-  zero argument method
+- Fix thread-safety issue in concurrent calls to zero-arg method in unmemoized
+  state which resulted in a `nil` value being accidentally returned in one thread
 
 ## [1.2.0] - 2021-11-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-- Nothing yet!
+- Fix theoretical thread-safety issue for concurrent calls to an uncached
+  zero argument method
 
 ## [1.2.0] - 2021-11-10
 
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -187,12 +187,11 @@ module MemoWise
           # hash key is just the method name.
           klass.module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
             def #{method_name}
-              _memo_wise_output = @_memo_wise[#{index}]
-              if _memo_wise_output || @_memo_wise_sentinels[#{index}]
-                _memo_wise_output
+              if @_memo_wise_sentinels[#{index}]
+                @_memo_wise[#{index}]
               else
-                @_memo_wise_sentinels[#{index}] = true
                 @_memo_wise[#{index}] = #{original_memo_wised_name}
+                @_memo_wise_sentinels[#{index}] = true
               end
             end
           END_OF_METHOD
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -192,6 +192,11 @@ module MemoWise
               else
                 ret = @_memo_wise[#{index}] = #{original_memo_wised_name}
                 @_memo_wise_sentinels[#{index}] = true
+                #unless frozen?
+                #  define_singleton_method(:#{method_name}) do
+                #    ret
+                #  end
+                #end
                 ret
               end
             end
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -192,11 +192,6 @@ module MemoWise
               else
                 ret = @_memo_wise[#{index}] = #{original_memo_wised_name}
                 @_memo_wise_sentinels[#{index}] = true
-                #unless frozen?
-                #  define_singleton_method(:#{method_name}) do
-                #    ret
-                #  end
-                #end
                 ret
               end
             end
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -190,8 +190,9 @@ module MemoWise
               if @_memo_wise_sentinels[#{index}]
                 @_memo_wise[#{index}]
               else
-                @_memo_wise[#{index}] = #{original_memo_wised_name}
+                ret = @_memo_wise[#{index}] = #{original_memo_wised_name}
                 @_memo_wise_sentinels[#{index}] = true
+                ret
               end
             end
           END_OF_METHOD
--- a/README.md
+++ b/README.md
@@ -117,15 +117,15 @@ Results using Ruby 3.0.2:
 
 |Method arguments|`Dry::Core`\* (0.7.1)|`Memery` (1.4.0)|
 |--|--|--|
-|`()` (none)|1.51x|19.82x|
-|`(a)`|2.30x|11.38x|
-|`(a, b)`|0.45x|2.10x|
-|`(a:)`|2.20x|22.83x|
-|`(a:, b:)`|0.49x|4.53x|
-|`(a, b:)`|0.46x|4.35x|
-|`(a, *args)`|0.89x|2.03x|
-|`(a:, **kwargs)`|0.82x|3.18x|
-|`(a, *args, b:, **kwargs)`|0.60x|1.62x|
+|`()` (none)|1.42x|17.84x|
+|`(a)`|2.48x|11.48x|
+|`(a, b)`|0.46x|2.05x|
+|`(a:)`|2.18x|20.50x|
+|`(a:, b:)`|0.48x|4.22x|
+|`(a, b:)`|0.48x|4.08x|
+|`(a, *args)`|0.90x|1.97x|
+|`(a:, **kwargs)`|0.80x|3.02x|
+|`(a, *args, b:, **kwargs)`|0.61x|1.54x|
 
 \* `Dry::Core`
 [may cause incorrect behavior caused by hash collisions](https://github.com/dry-rb/dry-core/issues/63).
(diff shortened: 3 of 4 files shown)


ms-ati:
Hi @jeremyevans! It is an honor to receive a PR from you - thank you for the `sequel` gem, among all your other impacts.

In order to work through this, as you mention in your PR above, we want to make sure we are making intentional choices to minimize performance regression (by benchmarking) while also ensuring we are fixing the thread-safety issue identified (through testing).

Do you have any thoughts you'd be able to share about minimal test case setup that can somewhat reliably manifest the race condition(s) you are identifying? I'm happy to help write up the test cases myself, if we can align on the best approach to structuring them.

Are you open to a bit of high-level test design and interacting on this? We also appreciate your help if you'd rather just do a drive-by PR of course.

ms-ati:
Also noting -- this change as written breaks the setting of `nil` as a return value -- but assuming we can work through a fix for that

ms-ati:
@jeremyevans if you have time, here's an example of a draft PR to try to test the race condition you raised:
https://github.com/panorama-ed/memo_wise/pull/225/files

Note that, as written, I can't surface the race condition in 1000 iterations of 2 threads on any tested Ruby version.

Is there a chance that, on existing Ruby versions, this race condition can't be reproduced in practice, as there is no Thread passing between the lines you identified on MRI nor JRuby?

jeremyevans:
This race condition should be very unlikely to hit, because it can only happen if the value is not yet cached.  Once the value has been cached, it's not possible to hit the race condition (no matter how many iterations you run).  I'm not sure you'll be able to force a reproduction on CRuby without introducing an explicit `Thread.pass` call (or something similar).

I'll see if I can clone the repository and make sure a cached nil value works correctly during the 3pm break today.  I just noticed this issue on the slide during the presentation and tried to fix it right away before I forgot about it.

jeremyevans:
I figured out what the problem was in the original patch.  I wasn't returning the correct value with the untested call.  I've pushed a new commit to fix that, and it does pass all specs.

I'm probably not in a good position to run/update benchmarks, since I run a fairly niche operating system (OpenBSD), on around 2014-era hardware.  I expect a slight decrease in performance.  If you want to improve performance, what you can do is after the first time a zero-arg method is called is to define a new method that returns the result directly.  With this approach, you don't even need a memo_wise cache.  Maybe I can send a commit with a different approach for that.

jeremyevans:
I've pushed a commit for a commented out version of a more optimized approach by defining a singleton method.  I assume this will improve performance, though I haven't benchmarked it.  It's commented out because it breaks tests. I think the tests are breaking because they are testing the implementation instead of the interface.  Let me know what you think.

ms-ati:
@jeremyevans unfortunately we have support for frozen objects as a design goal, so defining new methods at runtime to represented memoized zero-arg methods is something we ruled out for that reason.

jeremyevans:
FWIW, the commented out code in the last commit already only defines the singleton method if the object is not frozen.  There are potential downsides to defining singleton methods, such as creating singleton classes, so you would need to benchmark the performance difference and decide whether it is worth it.

lib/memo_wise.rb
```diff
@@ -187,12 +187,17 @@ def klass.extended(base)
           # hash key is just the method name.
           klass.module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
             def #{method_name}
-              _memo_wise_output = @_memo_wise[#{index}]
-              if _memo_wise_output || @_memo_wise_sentinels[#{index}]
-                _memo_wise_output
+              if @_memo_wise_sentinels[#{index}]
```
JacobEvelyn, line 190: Since this change will impact performance slightly, would you mind updating [the `README.md`'s benchmarks](https://github.com/panorama-ed/memo_wise#benchmarks) to use the numbers from [our CI run on your branch](https://github.com/panorama-ed/memo_wise/runs/4215740857?check_suite_focus=true)? For convenience, here they are:

```markdown
|Method arguments|`Dry::Core` (0.7.1)|`Memery` (1.4.0)|
|--|--|--|
|`()` (none)|1.42x|17.84x|
|`(a)`|2.48x|11.48x|
|`(a, b)`|0.46x|2.05x|
|`(a:)`|2.18x|20.50x|
|`(a:, b:)`|0.48x|4.22x|
|`(a, b:)`|0.48x|4.08x|
|`(a, *args)`|0.90x|1.97x|
|`(a:, **kwargs)`|0.80x|3.02x|
|`(a, *args, b:, **kwargs)`|0.61x|1.54x|
```

```markdown
|Method arguments|`DDMemoize` (1.0.0)|`Memoist` (0.16.2)|`Memoized` (1.0.2)|`Memoizer` (1.0.3)|
|--|--|--|--|--|
|`()` (none)|33.90x|3.44x|1.56x|4.03x|
|`(a)`|24.56x|17.02x|12.94x|14.91x|
|`(a, b)`|3.14x|2.35x|1.84x|2.03x|
|`(a:)`|34.42x|28.14x|23.83x|25.26x|
|`(a:, b:)`|5.13x|4.28x|3.77x|3.97x|
|`(a, b:)`|4.83x|4.08x|3.50x|3.66x|
|`(a, *args)`|3.03x|2.25x|1.95x|1.95x|
|`(a:, **kwargs)`|2.90x|2.44x|2.14x|2.24x|
|`(a, *args, b:, **kwargs)`|2.05x|1.77x|1.65x|1.64x|
```

ms-ati:
@JacobEvelyn Can we pause a moment before merging? I believe I have deterministic demonstration of the identified race condition even on MRI here: https://github.com/panorama-ed/memo_wise/pull/225

Apologies, my initial attempt contained a classed "accidental serialization" of the two competing threads, but with that fixed and having explicit `Thread.pass` in the zero-arg memoized function, it fails even on MRI.

If possible, my preference would be to align on the specifics of this **test case of the race condition**, and ideally to merge it separately as a **pending spec**, before we merge this fix -- so we have a standing record of the the problem and its demonstrated fix.

Thoughts on that approach?

lib/memo_wise.rb
```diff
@@ -187,12 +187,17 @@ def klass.extended(base)
...
+                @_memo_wise[#{index}]
               else
+                ret = @_memo_wise[#{index}] = #{original_memo_wised_name}
                 @_memo_wise_sentinels[#{index}] = true
-                @_memo_wise[#{index}] = #{original_memo_wised_name}
+                #unless frozen?
```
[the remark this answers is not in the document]
jeremyevans: Reverted the last commit.

lib/memo_wise.rb
jeremyevans: Updated README with these numbers.

lib/memo_wise.rb
```diff
@@ -187,12 +187,17 @@ def klass.extended(base)
...
+                #  define_singleton_method(:#{method_name}) do
+                #    ret
+                #  end
+                #end
+                ret
```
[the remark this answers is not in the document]
jeremyevans: Updated CHANGELOG.

jeremyevans:
> Hi @jeremyevans ! Thanks so much for the PR, and I enjoyed your talk as well! 💞

Thank you. :)

> I think the `define_singleton_method` approach breaks tests because we also support the ability to preset/reset memoized values (which we did not mention in the talk), and our implementation of those features relies on the `@_memo_wise` and `@_memo_wise_sentinels` data structures being used in the memoized method.

Ah, that makes sense.  If you are allowing resets, defining singleton methods wouldn't work unless you were keeping track of the defined singleton methods, and then on cache reset, removing them from the singleton class with `remove_method`.

> We may be able to implement those features in a different way to still use `define_singleton_method`—and welcome your ideas on that or any other performance optimizations to pursue—but in the meantime I think it makes sense to remove that commented-out conditional to make sure our code is correct.

Another way to get this to work is to override `freeze`, and before calling `super`, create singleton methods for all cached zero-argument methods.  Then cache resets don't matter, since I assume you don't allow resets for frozen objects.  I use this approach in other libraries, to improve performance when freezing objects or classes.

> @JacobEvelyn Can we pause a moment before merging? I believe I have deterministic demonstration of the identified race condition even on MRI here: #225
> 
> Apologies, my initial attempt at the test code contained a classic "accidental serialization" of the two competing threads, meaning that they didn't actually run concurrently. Now that this is fixed, and having explicit `Thread.pass` in the zero-arg memoized function itself, the test is now failing reliably even on MRI, by triggering the identified race condition.
> 
> My preference would be to align (ideally including @jeremyevans if you have thoughts) on the specifics of the **test case of the race condition**, and merge the test as a **pending spec**, before we go ahead and merge this fix. That way, we will have a standing record of the the problem and its demonstrated fix, as well as a regression test.
> 
> Thoughts on that approach?

That sounds like a fine approach to me.  Thank you for taking the time to put together a working spec.  I know from experience how hard it is to do that.

ms-ati:
@jeremyevans Ok, the pending spec for `thread_safety_spec.rb` was merged in #225 -- looking forward to rebasing this fix on top and fix the pending spec 😄

jeremyevans:
OK, I rebased against main and removed the pending marker from spec.

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-- Nothing yet!
+- Fix theoretical thread-safety issue for concurrent calls to an uncached
+  zero argument method
```
ms-ati, line 10:
```suggestion
- Fix thread-safety issue in concurrent calls to zero-arg method in unmemoized
  state which resulted in a `nil` value being accidentally returned in one thread
```

ms-ati: Apologies for the nit-pick, but I would say this issue was demonstrably _not purely theoretical_ 😄

ms-ati (approved):
LGTM!

JacobEvelyn (approved):

JacobEvelyn:
Thanks for your patience @jeremyevans !

Issue #184, opened 2021-07-12
Question: Any test/doc about thread safety when using this gem?


PikachuEXE:
I shared this gem on Reddit and someone ask for thread safety:
https://old.reddit.com/r/rails/comments/oglovy/introducing_memowise/

I am not very concerned for my own use but would be good to have something for it.

ms-ati:
Thread-safety of **zero-args** methods
- Pending spec in #225 
- Fix for thread-safety in #224 
- TODO: Document and add spec demonstrating that an unmemoized method may be called multiple times under contention  between threads, and one of the return values will be set as the memoized value

PR #225, opened 2021-11-10
Add spec of thread-safety accessing zero-arg method

ms-ati:
See discussion in #224

### DIFF
--- a/spec/thread_safety_spec.rb
+++ b/spec/thread_safety_spec.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+RSpec.describe "thread safety" do # rubocop:disable RSpec/DescribeClass
+  context "when two threads accessing unmemoized zero-args method" do
+    let(:thread_return_values) do
+      any_nils = ->(a) { a.any?(&:nil?) }
+
+      check_repeatedly(condition_proc: any_nils) do
+        instance = class_with_memo.new
+        threads = Array.new(2) { Thread.new { instance.first_thread_id } }
+        threads.map(&:value)
+      end
+    end
+
+    let(:class_with_memo) do
+      Class.new do
+        prepend MemoWise
+
+        def first_thread_id
+          Thread.pass # trigger a race condition even on MRI
+          Thread.current.object_id
+        end
+        memo_wise :first_thread_id
+      end
+    end
+
+    # Tip of the hat to @jeremyevans for finding and fixing a race condition
+    # which caused accidental `nil` values to be returned under contended calls
+    # to unmemoized zero-args methods.
+    #   * See https://github.com/panorama-ed/memo_wise/pull/224
+    it "does not return accidental nil value to either thread" do
+      pending("fix for zero-arg race condition in #224")
+      expect(thread_return_values).not_to include(nil)
+    end
+  end
+end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -46,6 +46,7 @@ require "bundler/setup"
 require "memo_wise"
 
 # Support code for tests
+require "support/check_repeatedly"
 require "support/define_methods_for_testing_memo_wise"
 require "support/shared_context_for_instance_methods"
 require "support/shared_context_for_class_methods_via_self_dot"
--- a/spec/support/check_repeatedly.rb
+++ b/spec/support/check_repeatedly.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+# Helper method repeatedly checks for a probabilistically occurring condition
+# by running the given block until either:
+#   (a) given condition is detected, or
+#   (b) max number of seconds have elapsed.
+#
+# When adding tests for race conditions, we choose to implement time-limited
+# probabilistic test cases, so that rarely-occurring cases might be detected
+# over time in CI runs, without slowing down a single run of the test suite
+# too much.
+#
+# @param condition_proc [Proc]
+#   Condition proc to check against yielded values, will stop repeating and
+#   return if this returns `true`.
+#
+# @param up_to_max_secs [Numeric]
+#   Max number of seconds to run, default: 1.
+#
+# @yieldreturn [Object]
+#   Return value of given block is checked against given condition on each
+#   iteration. Final value will be returned to caller.
+#
+# @return [Object]
+#   Returns the final return value of the given block.
+def check_repeatedly(condition_proc:, up_to_max_secs: 1.0)
+  time_max = Time.now.utc + up_to_max_secs
+
+  loop do
+    value = yield
+    return value if condition_proc.call(value) || Time.now.utc > time_max
+  end
+end


ms-ati:
cc @JacobEvelyn @jemmaissroff this is a pending spec for the thread-safety issue identified and fixed by @jeremyevans in #224 -- appreciate your review, so we can merge this pending spec before the fix?

spec/thread_safety_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  context "when two threads accessing unmemoized zero-args method" do
+    # Tip of the hat to @jeremyevans,
+    # see https://github.com/panorama-ed/memo_wise/pull/224
+    let(:thread_return_values) do
+      thread_return_values = [nil, nil]
+
+      1000.times do
+        instance = class_with_memo.new
+
+        threads = Array.new(2).map do
```
jemmaissroff, line 13: I think the `map` is superfluous with `Array.new(2)` as well. `Array#new` can optionally take a block, so we could do:

```ruby
Array.new(2) do
   ...
end
```

spec/thread_safety_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+          Thread.new { instance.return_first_thread_id }
+        end
+
+        thread_return_values = threads.map(&:value)
```
jemmaissroff, line 17: Do we need to define `threads` at all, or could the above become:

```ruby
thread_return_values = Array.new(2) do
  Thread.new { instance.return_first_thread_id }
end.map(&:value)
```

spec/thread_safety_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+RSpec.describe "thread safety" do # rubocop:disable RSpec/DescribeClass
+  context "when two threads accessing unmemoized zero-args method" do
+    # Tip of the hat to @jeremyevans,
+    # see https://github.com/panorama-ed/memo_wise/pull/224
+    let(:thread_return_values) do
```
jemmaissroff, line 7: I think this whole block could be simplified to

```ruby
let(:thread_return_values) do
  instance = class_with_memo.new

  Array.new(2) do
    Thread.new { instance.return_first_thread_id }
  end.map(&:value)
end
```

Rationale:
* Now that it's deterministic, we don't need the `1000.times` block
* `Array#new` takes an optional block, so we don't need the `#map` there too
* We don't need the break condition anymore
* No need to define `threads` or `thread_return_values`

spec/thread_safety_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+          Thread.pass # trigger a race condition even on MRI
+          Thread.current.object_id
+        end
+        memo_wise :return_first_thread_id
+      end
+    end
+
+    # TODO: Pending proposed fix in #224, remove this comment and enable spec
+    pending "returns the same memoized non-nil thread id in each thread" do
+      expect(thread_return_values[0]).to eq(thread_return_values[1])
```
jemmaissroff, line 42: I think this test will still fail even with @jeremyevans's fix. The rationale is that both threads will still hit this line:
```ruby
ret = @_memo_wise[#{index}] = #{original_memo_wised_name}
```
at roughly the same time, and still therefore get different values from their respective `return_first_thread_id` calls.

The case that Jeremy's PR has fixed is the nil case - neither will have a `nil` value. Do we also want to address this first case and ensure that both threads have the same memoized values? If so, might we need to use a Mutex or something like that (presumably at a performance cost in the non-memoized case) to lock the `@_memo_wise` cache?

jemmaissroff (changes requested):

spec/thread_safety_spec.rb
jeremyevans: I think it's expected that any method that is cached should return the same value no matter which thread is calling it.  The test breaks that assumption.  Maybe the test should just check that the value returned is not nil, because in the failure case before my fix, `nil` should be returned if the race is lost.

spec/thread_safety_spec.rb
ms-ati: @jemmaissroff while it is true that in the current code and test, we can indeed reliably reproduce the race on the first iteration, are you sure it's best to leave that assumption in place for an on-going regression test? In general, when leaving tests for previous race conditions in code, my norm has been to leave enough iterations up to 1 or 2 seconds of test time, so that if we every have a case that occurs probabilistically, which is the norm, we will eventually catch it in CI.

I'm open to thoughts either way! I just don't want my "success" here in triggering on the first iteration to have too much influence in how we think about designing race condition regression tests, if that makes sense.

spec/thread_safety_spec.rb
ms-ati: 😝  I was hoping nobody would see that until we went to enable this test on Jeremy's PR

ms-ati: (I was hoping to drive an explicit decision on this point within the context of his proposed fix)

What if I split the assertions into two, separate pending specs? And we can discuss whether Jeremy's PR should close one of them, and the other can be addressed in a separate PR, or converted to a documentary spec that shows that we expect possibly multiple values when the memoization is contended? (which to be clear is an acceptable design decision, that we simply need to document if so)

ms-ati: @jeremyevans looks like we "typed past each other" here, as you can see in the timestamps of our comments above ;)

So to playback what we are all saying: this test has two distinct expectations.

Split-tracking:

- (a) Test of accidental `nil`
  - Proposal: Separate to its own test + remove `pending` as part of Jeremy's PR that fixes it?

- (b) Test of same return value when the memoization is contended
  - Option 1: Remove this test entirely
  - Option 2: Change test to be documentary - that this issue exists and should be expected
  - Option 3: Separate `pending` test - and plan to introduce a separate fix per Jeremy's description of the expectation above that we should not return different values when memoization is contended

Re-reading what everyone wrote above, what do we think about (a) separating, and (b-3) leaving a separate pending test in place to address the contended values issue in a different PR?

jeremyevans: I personally see no value in testing behavior for a non-idempotent method.  Nobody should be using caching for non-idempotent methods.  So I would only check that the race condition is fixed (the test for non-nil results on a method that returns non-nil).

ms-ati: I am totally fine removing this case entirely for now. 

I suspect that, per #184, we must eventually document what happens in contended cases, for example:

- Queries of external state
  - Should reading data in from a web request, database query, or file be memoized?
  - If so, what is the actual vs expected behavior, when contended queries generate different query results?

Unless I hear back differently from @jemmaissroff or @JacobEvelyn, I will plan to remove this test case for now, and we can discuss under #184.

JacobEvelyn: Thanks for the thoughts, @ms-ati @jeremyevans @jemmaissroff ! I don't have strong feelings on what the behavior should be for nondeterministic methods, so I'm happy to defer to you three on either providing a stronger consistency guarantee or simply changing the test to confirm that the results are not `nil`. I agree with @ms-ati that we should document the behavior wherever we land.

spec/thread_safety_spec.rb
JacobEvelyn: I don't think I have strong feelings here. I see the argument for letting this run for a while to catch nondeterministic failures, but I also think @jemmaissroff 's point about simplifying this if the failure _is_ deterministic makes sense. I'd be happy with either:

- simplifying the test as @jemmaissroff suggests
- keeping the code as-is and adding a comment explaining the rationale that @ms-ati outlined above

ms-ati: Request for re-review @jemmaissroff -- I've extracted code and added comments to provide a rationale for time-limited probabilistic testing strategy for race conditions. What do you think?

spec/thread_safety_spec.rb
ms-ati: Per discussion above, I've removed the test of same return value under contended memoization, we can discuss this separately.

spec/thread_safety_spec.rb
jemmaissroff: Since this race condition reproduction is deterministic, I personally was confused by the repetition here.

Although I can admire the elegance of `check_repeatedly`, it seems to me to be over-engineering in this case and solving for a problem that we don't currently have. If we at some future point either have a race condition which we need to repeatedly run to reproduce, or this race condition for some reason requires repetition, this method looks perfect.

In the absence of this however, I think the repetition here overcomplicates and makes it harder to reason about a test which only requires one iteration to accomplish its goal.

ms-ati: @jemmaissroff Ok, it seems like you would advocate for not having the practice for time-limited probabilistic testing in place until we have discovered a race condition which manifests probabilitically?

I thought having this in place in this new file `thread_safety.rb` would lower the work required for us to add tests for other race conditions, as we work to land thread-safety and documentation of thread contention behavior across the project.

So is there a natural tension between these interests?

1. Avoid engineering test code for probabilistic race conditions until we have one at hand
2. Make it easier to detect and add tests for probabilistic race conditions

jemmaissroff: I might be missing context here - do we have plans to add any other race condition tests? Or to do more work on creating more race condition tests?

ms-ati: _To be clear: I am happy removing this machinery for now!_

I am hoping to reach alignment on whether there are distinct interests at play in this decision? Are we choosing to serve 1 over 2, or are saying that 2 is not a valid interest ("over-engineering" and "overcomplicates")?

jemmaissroff: Yes, I am aligned with removing it.

I'm trying to understand whether 2 is a valid interest. My assumption earlier in this thread was that we don't have plans to add other race condition tests, and so under that assumption I suggested that it is overcomplicating our codebase. When you outlined the interests, it made me realize I had made this assumption, and so I'm looking for clarity on whether we do have plans to add more race condition tests or not. 

If we don't have plans to do this, I would say yes, accommodating a case that doesn't exist is not a valid interest.

If we do have plans to do this, I'm interested in hearing more about this before I feel equipped to judge which interest we should prioritize.

ms-ati: Ah ok, thanks for clarifying @jemmaissroff. I had been under the assumption that this was the beginning of a series of work to eventually fully answer #184 -- and I had been imagining adding test coverage for potential races across all of the memoization code paths.

But I find it very persuasive that such tooling can wait at least one more PR, since this is the very first, none of those other plans might happen, and it isn't needed for just this PR.

I'll remove it for now and we can revisit if we do end up moving forward with thread-safety tests across all of the memoization paths.

ms-ati: @jemmaissroff Removed as discussed, please re-review?

ms-ati: Well 😆 the joke is on us @jemmaissroff @JacobEvelyn -- in fact, it appears that the issue is not actually deterministic, just appears often. So in fact, trying it once is not enough, as you can see from these CI failures: https://github.com/panorama-ed/memo_wise/runs/4231683800?check_suite_focus=true

Q: Shall I return the machinery for time-limited probabilistic tests then?

spec/thread_safety_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        threads = Array.new(2) { Thread.new { instance.return_first_thread_id } }
+        threads.map(&:value)
+      end
+    end
+
+    let(:class_with_memo) do
+      Class.new do
+        prepend MemoWise
+
+        def return_first_thread_id
```
JacobEvelyn, line 19: Nitpick: I wonder if a more "Rubyesque" name would just be `first_thread_id`.

JacobEvelyn (approved):
LGTM with one optional change. Would like @jemmaissroff 's review on this too before merging.

spec/thread_safety_spec.rb
ms-ati: Done!

jemmaissroff (approved):

PR #229, opened 2021-11-18
Fix a variety of inheritance bugs

### LINKED ISSUE #228: Broken handling of inheritance after "Optimize zero-argument methods"
Seem that after d768e1091ae112531c4bf34534f43aa1f37c9761 inheritance is not handled properly, because used indices are mixed up between classes.

```ruby
require 'memo_wise'

class Parent
  prepend MemoWise

  def bar
    'bar'
  end
  memo_wise :bar
end

class Child < Parent
  def foo
    'foo'
  end
  memo_wise :foo
end

child = Child.new
pp child.foo # initialize Child sentinels
pp child.bar # expected 'bar', got 'foo' because Child sentinels used instead of Parent
```

JacobEvelyn:
This commit fixes a variety of bugs that occurred when a
parent class uses `MemoWise` and a child class inherits
from it and tries to use the memoized methods.

Fixes #228

**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -156,6 +156,17 @@ module MemoWise
           klass = klass.singleton_class
         end
 
+        if klass.singleton_class?
+          # This ensures that a memoized method defined on a parent class can
+          # still be used in a child class.
+          klass.module_eval <<-END_OF_METHOD, __FILE__, __LINE__ + 1
+            def inherited(subclass)
+              super
+              MemoWise::InternalAPI.create_memo_wise_state!(subclass)
+            end
+          END_OF_METHOD
+        end
+
         raise ArgumentError, "#{method_name.inspect} must be a Symbol" unless method_name.is_a?(Symbol)
 
         api = MemoWise::InternalAPI.new(klass)
@@ -167,19 +178,7 @@ module MemoWise
         klass.send(:private, original_memo_wised_name)
 
         method_arguments = MemoWise::InternalAPI.method_arguments(method)
-        # `@_memo_wise_indices` stores the `@_memo_wise` indices of different
-        # method names. We only use this data structure when resetting or
-        # presetting memoization. It looks like:
-        #   {
-        #     single_arg_method_name: 0,
-        #     other_single_arg_method_name: 1
-        #   }
-        memo_wise_indices = klass.instance_variable_get(:@_memo_wise_indices)
-        memo_wise_indices ||= klass.instance_variable_set(:@_memo_wise_indices, {})
-        index = klass.instance_variable_get(:@_memo_wise_index_counter) || 0
-
-        memo_wise_indices[method_name] = index
-        klass.instance_variable_set(:@_memo_wise_index_counter, index + 1)
+        index = MemoWise::InternalAPI.next_index!(klass, method_name)
 
         case method_arguments
         when MemoWise::InternalAPI::NONE
--- a/lib/memo_wise/internal_api.rb
+++ b/lib/memo_wise/internal_api.rb
@@ -161,13 +161,70 @@ module MemoWise
     def self.original_class_from_singleton(klass)
       raise ArgumentError, "Must be a singleton class: #{klass.inspect}" unless klass.singleton_class?
 
+      # Since we call this method a lot, we memoize the results. This can have a
+      # huge impact; for example, in our test suite this drops our test times
+      # from over five minutes to just a few seconds.
+      @original_class_from_singleton ||= {}
+
       # Search ObjectSpace
       #   * 1:1 relationship of singleton class to original class is documented
       #   * Performance concern: searches all Class objects
-      #     But, only runs at load time
-      ObjectSpace.each_object(Module).find do |cls|
+      #     But, only runs at load time and results are memoized
+      @original_class_from_singleton[klass] ||= ObjectSpace.each_object(Module).find do |cls|
         cls.singleton_class == klass
       end
+    end
+
+    # Increment the class's method index counter, and return an index to use for
+    # the given method name.
+    #
+    # @param klass [Class]
+    #   Original class on which a method is being memoized
+    #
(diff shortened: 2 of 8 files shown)


lib/memo_wise.rb
```diff
@@ -176,10 +183,26 @@ def klass.extended(base)
...
+        # However, we don't use a class variable for `@_memo_wise_indices`
+        # because we want to allow instance and class methods with the same name
+        # to both be memoized, and using a class variable would share that index
+        # data between them.
+        counter_class = klass.singleton_class? ? MemoWise::InternalAPI.original_class_from_singleton(klass) : klass
+        index = if counter_class.class_variable_defined?(:@@_memo_wise_index_counter)
+                  counter_class.class_variable_get(:@@_memo_wise_index_counter)
+                else
+                  0
+                end
```
ms-ati, line 202: Is this section a candidate for extraction to `InternalAPI`?

lib/memo_wise.rb
```diff
@@ -243,7 +238,14 @@ def #{method_name}(#{MemoWise::InternalAPI.args_str(method)})
       end
     end
 
-    unless target.singleton_class?
+    if target.singleton_class?
+      # This ensures that a memoized method defined on a parent class's
+      # singleton class can still be used by a child class.
+      def (MemoWise::InternalAPI.original_class_from_singleton(target)).inherited(subclass) # rubocop:disable Lint/NestedMethodDefinition
```
JacobEvelyn, line 244: One thing I don't fully understand—tests pass when I change this to just `def inherited(subclass)` (except the one test that checks which `public_instance_methods` are added, so behavior _is different_). @jemmaissroff and I dug in a little bit but couldn't figure out why this approach works. @ms-ati do you know?

lib/memo_wise/internal_api.rb
```diff
@@ -161,15 +161,70 @@ def self.key_str(method)
...
+    #
+    # @param klass [Class]
+    #   Original class on which a method is being memoized
+    #
+    # @param method_name [Symbol]
+    #   The name of the method being memoized
+    #
+    # @return [Integer]
+    #   The index within `@_memo_wise` to store the method's memoized results
+    def self.index!(klass, method_name)
```
ms-ati, line 187: Wonder if this should have a more explicit indication of side-effect in the name? Something like `next_index!` or `inc_index!` or `succ_index!`? Just an idea to consider.

lib/memo_wise.rb
```diff
@@ -110,6 +110,13 @@ def allocate
         MemoWise::InternalAPI.create_memo_wise_state!(super)
       end
 
+      # This ensures that a memoized method defined on a parent class can still
+      # be used in a child class.
+      def inherited(subclass)
+        super
+        MemoWise::InternalAPI.create_memo_wise_state!(subclass)
+      end
```
ms-ati, line 118: Should this be guarded in some way to make sure we need the MemoWise state on the class itself? I can't remember if we always create the class-level state, or only if we are memo_wise-ing class methods?

lib/memo_wise/internal_api.rb
```diff
@@ -161,15 +161,70 @@ def self.key_str(method)
...
+      # Since we call this method a lot, we memoize the results.
+      @original_class_from_singleton ||= {}
+
       # Search ObjectSpace
       #   * 1:1 relationship of singleton class to original class is documented
       #   * Performance concern: searches all Class objects
-      #     But, only runs at load time
-      ObjectSpace.each_object(Module).find do |cls|
+      #     But, only runs at load time and results are memoized
+      @original_class_from_singleton[klass] ||= ObjectSpace.each_object(Module).find do |cls|
```
ms-ati, line 171: To what extent is this additional state (which lives forever inside the `MemoWise` module) justified by quantifiable performance numbers?

JacobEvelyn: Great question! It's possible that real-world behavior will typically be less noticeable, but this has a huge impact in our tests. With this change:

```
Finished in 3.53 seconds (files took 0.51088 seconds to load)
438 examples, 0 failures
```

Without this memoization:

```
Finished in 5 minutes 25 seconds (files took 0.41548 seconds to load)
438 examples, 0 failures
```

Should I make that clearer in the comment?

lib/memo_wise.rb
JacobEvelyn: Great question. I think ideally we'd only want to do this when we need it, but I don't see a particularly easy path to that right now, and I see the same issue with `def initialize` and `def allocate`. Would it make sense to make a separate issue for that work?

ms-ati: Hmm I think they are different cases fundamentally.

- Initialize and allocate are **only** called on instances. They are not every triggered for classes, right?

lib/memo_wise/internal_api.rb
ms-ati: Yeah, I think perhaps in the comment would be helpful?

lib/memo_wise.rb
ms-ati: Could this be a bit more readable / accessible to readers as:

```suggestion
      parent_class = MemoWise::InternalAPI.original_class_from_singleton(target)
      def parent_class.inherited(subclass) # rubocop:disable Lint/NestedMethodDefinition
```

spec/adding_methods_spec.rb
```diff
@@ -44,5 +44,26 @@
...
+
+    context "when a class method is memoized" do
+      subject do
+        klass.send(:prepend, MemoWise)
+        klass.send(:memo_wise, self: :example)
+      end
+
+      let(:klass) do
+        Class.new do
+          def self.example ; end
```
ms-ati, line 56: Fix Rubocop?

```suggestion
          def self.example; end
```

ms-ati (approved):
LGTM! Thank you @JacobEvelyn !

PR #230, opened 2021-11-18
Fix concurrent memoization of zero-arg methods multiple values

ms-ati:
When two threads concurrently call a zero-arg method which is in
the unmemoized state, they can return two different results.

For example, this could occur if we are memoizing:
  * database query result
  * filesystem read
  * current time
  * a random number

This PR adds documentation that this is expected behavior to README.md,
as well as a probabilistic spec that demonstrates the behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,24 @@ $ bundle exec ruby benchmarks.rb
 If your results differ from what's posted here,
 [let us know](https://github.com/panorama-ed/memo_wise/issues/new)!
 
+## Thread Safety
+
+MemoWise makes the following **thread safety** guarantees on all supported Ruby
+versions:
+
+1. **Before** a value has been memoized
+
+   * Contended calls from multiple threads...
+      * May each call the original method
+      * May return different valid results (when the method is nondeterministic,
+        like `rand`)
+      * Will memoize exactly one valid return value
+
+2. **After** a value has been memoized
+
+   * Contended calls from multiple threads...
+     * Always return the same memoized value
+
 ## Documentation
 
 ### Documentation is Automatically Generated
--- a/spec/thread_safety_spec.rb
+++ b/spec/thread_safety_spec.rb
@@ -3,11 +3,9 @@
 RSpec.describe "thread safety" do # rubocop:disable RSpec/DescribeClass
   context "when two threads accessing unmemoized zero-args method" do
     let(:thread_return_values) do
-      any_nils = ->(a) { a.any?(&:nil?) }
-
-      check_repeatedly(condition_proc: any_nils) do
-        instance = class_with_memo.new
-        threads = Array.new(2) { Thread.new { instance.first_thread_id } }
+      check_repeatedly(condition_proc: condition_to_check) do
+        @instance = class_with_memo.new
+        threads = Array.new(2) { Thread.new { @instance.current_thread_id } } # rubocop:disable RSpec/InstanceVariable
         threads.map(&:value)
       end
     end
@@ -16,11 +14,11 @@ RSpec.describe "thread safety" do # rubocop:disable RSpec/DescribeClass
       Class.new do
         prepend MemoWise
 
-        def first_thread_id
-          Thread.pass # trigger a race condition even on MRI
-          Thread.current.object_id
+        def current_thread_id
+          Thread.pass              # trigger a race condition even on MRI
+          Thread.current.object_id # return different values in each thread
         end
-        memo_wise :first_thread_id
+        memo_wise :current_thread_id
       end
     end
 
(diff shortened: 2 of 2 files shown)


ms-ati:
As expected: this purely ruby code "double check" pattern, where we check again whether the memoization sentinel has been set in a concurrent thread _after_ we have run the original method, is successful in this test on all MRI versions, but does not work on JRuby.

**Options**

1. We could leave the code in its prior state, and document that contended calls can result in different return value across threads on all Ruby versions
2. We could adopt this solution for MRI, and document that it doesn't work on JRuby in both markdown and tests (e.g. by marking this test as "pending" in JRuby)
3. We might go a step further and use a Mutex on each object, which would fix this issue on all Ruby versions, at some additional cost on those first calls where we set the memoized value

I'm hoping that this PR can help kick off a discussion and we can intentionally select an option 😄

ms-ati:
Summary of discussion with @jemmaissroff and @JacobEvelyn outside of Github:

- We've selected **Option 1: leave code in is prior state, and document that contended calls can result in different return values across threads**

README.md
```diff
@@ -156,6 +156,23 @@ $ bundle exec ruby benchmarks.rb
...
 
+## Thread Safety
+
+MemoWise makes the following **thread safety** guarantees on all supported Ruby
+versions:
+
+1. **Before** a value has been memoized
+
+   * Contended calls from multiple threads...
+      * May call original method multiple times
```
JacobEvelyn, line 167: I was a bit confused by this wording. Maybe something like:

* May each call the original method

?

README.md
```diff
@@ -156,6 +156,23 @@ $ bundle exec ruby benchmarks.rb
...
+      * May return different valid return values on different threads
```
JacobEvelyn, line 168: I worry people may not understand the context of this one. Maybe something like:

* May return different valid results (when the method is nondeterministic, like `rand`)

spec/thread_safety_spec.rb
```diff
@@ -16,20 +14,65 @@
...
+
+      it "does not return accidental nil value to either thread" do
+        expect(thread_return_values).not_to include(nil)
+      end
+    end
+
+    # When memoizing non-deterministic methods, MemoWise does not impose any
+    # additional guarantees other than:
+    #   * Return values are one of the valid returns from the original method
+    #   * Return values differ between threads under contended calls
```
JacobEvelyn, line 42: This isn't a guarantee, right? Maybe we want _may_ differ or some other rephrasing?

spec/thread_safety_spec.rb
```diff
@@ -16,20 +14,65 @@
...
+    #   * One of those return values will be memoized and returned thereafter
+    context "when checking condition: are different values returned?" do
+      let(:condition_to_check) do
+        ->(values) { values.any? { |v| v != values.first } }
```
JacobEvelyn, line 46: This is fine to leave as-is but I found it a little confusing and am wondering if `values.uniq.size > 1` would be clearer?

spec/thread_safety_spec.rb
```diff
@@ -16,20 +14,65 @@
...
+      end
+
+      let(:after_memoization_thread_return_values) do
+        _ = thread_return_values # sanity check - always comes after first `let`
```
JacobEvelyn, line 50: I find this line a little confusing. Do we need the `_ = `? Would it be simpler to just make `thread_return_values` a `let!`? Would the comment be clearer as something like:

```ruby
thread_return_values # Ensure threads have already executed
```

?

spec/thread_safety_spec.rb
```diff
@@ -16,20 +14,65 @@
...
+        end
+      end
+
+      # NOTE: Disabled on TruffleRuby in CI, because it takes multiple seconds
+      # to observe the different values returned to multiple threads, and errors
+      # out in some cases when that does happen.
+      unless RUBY_ENGINE == "truffleruby" && ENV["CI"] == "true"
+        it "returns different values to each thread, and memoizes one of them" do
+          # Before memoization: expect to observe threads return different values
+          expect(thread_return_values).
```
JacobEvelyn, line 64: Similar to my comment above, would this be clearer?

```ruby
expect(thread_return_values.uniq.size).to be > 1
```

I found the `include(not_eq(... .first))` difficult to understand.

README.md
```diff
@@ -156,6 +156,23 @@ $ bundle exec ruby benchmarks.rb
...
+      * Will memoize exactly one valid return value
+
+2. **After** a value has been memoized
+
+   * Contended calls from multiple threads...
+     * Always returns the same memoized value
```
JacobEvelyn, line 174: Nitpick: `returns` --> `return`

JacobEvelyn (commented):
Just some minor comments. Thanks for working on this!

Also, consider adding this to the bottom of your commit message:

```
Closes #184 
```

README.md
ms-ati: Changed!

spec/thread_safety_spec.rb
ms-ati: Changed! Thanks

spec/thread_safety_spec.rb
ms-ati: Changed, thanks

spec/thread_safety_spec.rb
ms-ati: Changed, thanks!

JacobEvelyn (approved):

PR #233, opened 2021-11-23
Switch heredocs from <<- to <<~

### LINKED ISSUE #232: Use squiggly heredocs everywhere
Squiggly heredocs seem like the "more correct" way to use multiline strings for our use case, and they're supported in all Ruby versions we support, so we should switch to using them everywhere.

JacobEvelyn:
This commit also changes the heredoc closing
signifier to `HEREDOC` everywhere, for clarity.

Closes #232

**Before merging:**

### DIFF
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -63,8 +63,7 @@ BENCHMARK_GEMS = [
 # Use metaprogramming to ensure that each class is created in exactly the
 # the same way.
 BENCHMARK_GEMS.each do |benchmark_gem|
-  # rubocop:disable Security/Eval
-  eval <<-CLASS, binding, __FILE__, __LINE__ + 1
+  eval <<~HEREDOC, binding, __FILE__, __LINE__ + 1 # rubocop:disable Security/Eval
     # For these methods, we alternately return truthy and falsey values in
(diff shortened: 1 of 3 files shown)


benchmarks/benchmarks.rb
```diff
@@ -63,8 +63,7 @@ def benchmark_name
 # Use metaprogramming to ensure that each class is created in exactly the
 # the same way.
 BENCHMARK_GEMS.each do |benchmark_gem|
-  # rubocop:disable Security/Eval
-  eval <<-CLASS, binding, __FILE__, __LINE__ + 1
+  eval <<~HEREDOC, binding, __FILE__, __LINE__ + 1 # rubocop:disable Security/Eval
```
JacobEvelyn, line 66: I don't love `HEREDOC` but was wondering if we might want something more consistent. Right now we use `CLASS`, `END_OF_METHOD`, and `END_OF_METHODS`. I'm happy to revert this change though and keep them as-is, or switch to something more understandable than `HEREDOC`, like `MULTILINE_STRING`.

Thoughts?

ms-ati: I am fine with whatever you choose here @JacobEvelyn. Agree that having a consistent usage might be nice.

ms-ati (approved):
LGTM

PR #246, opened 2021-12-09
Implements memoization using instance variables per method

### LINKED ISSUE #243: Use method-specific instance variables for methods that take arguments
This was a suggestion made by @sampersand in the RubyConf Discord channel for our recent talk. The idea is that instead of using `@_memo_wise` for all method types, we'd use a different instance variable for each method, like `@_memo_wise_method_#{method_name}`. This should save us an array lookup for a slight performance boost.

There are a few challenges:

- we'd need to avoid name collisions, for example when memoizing both a `data?` and a `data!` and a `data` method
  - probably a simple `.sub("?", "__qmark__").sub("!", "__bang__")` would be sufficient
  - @jemmaissroff may have insight into whether there's a max instance variable length to be worried about (or a length at which performance decreases)
    - if so, we could instead use a scheme like `@_memo_wise_method_#{counter}` or use UUIDs, etc.
- to support frozen objects we need to ensure that these instance variables are initialized to empty hashes before freezing
  - this could probably be done either in an overridden `initialize` or `freeze` method
- this approach will not support resetting and presetting zero-arity methods on frozen objects
  - a simple path forward would be to continue using our array for zero-arity methods
  - are there good alternatives?

Would love discussion on this idea and its tradeoffs!

jemmaissroff:
Resolves #243

**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -180,6 +180,7 @@ module MemoWise
 
         case method_arguments
         when MemoWise::InternalAPI::NONE
+          index = MemoWise::InternalAPI.index(klass, method_name)
           # Zero-arg methods can use simpler/more performant logic because the
           # hash key is just the method name.
           klass.send(:define_method, method_name) do # Ruby 2.4's `define_method` is private in some cases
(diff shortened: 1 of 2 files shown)


PikachuEXE:
CI for 3.0 failed for `Branch coverage (92.59%) is below the expected minimum coverage (100.00%).`
But
https://codecov.io/github/panorama-ed/memo_wise/commit/4c549d4302e3164f8211e8f15cffe96ae5d2f7e9
shows 100%

I can't understand this 😅

PikachuEXE:
Oh wait there is code conflict

jemmaissroff:
Not planning on working on this in the immediate future

PR #247, opened 2021-12-09
[WIP] Change MemoWise public API

paracycle:
Re-attempt at #213 with @jemmaissroff, still a WIP...

**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Nothing yet!
+- Switch to a simpler internal data structure to fix several classes of bugs
+  that the previous few versions were unable to sufficiently address
 
 ## [1.4.0] - 2021-12-10
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Or install it yourself as:
 
 ## Usage
 
-When you `prepend MemoWise` within a class or module, `MemoWise` exposes three
+When you `extend MemoWise` within a class or module, `MemoWise` exposes three
 methods:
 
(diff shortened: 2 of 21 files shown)


PikachuEXE:
Should I wait for this to be finished before migrating to this gem?

paracycle:
> Should I wait for this to be finished before migrating to this gem?

I am not a maintainer of the gem, so I am not sure if this will ever get merged and released. But, my aim is to keep the public API surface change to only the change from `prepend MemoWise` to `extend MemoWise`. So, if you adopt now, it shouldn't be hard to upgrade later.

ms-ati:
Hi @PikachuEXE, my understanding is that this is intended to be a proposal to be discussed -- which is awesome! -- but at this time, there's no current understanding nor expectation that this proposed API change will be accepted, is there?

UPDATE @paracycle I see your response first, and 👍🏻 to how you expressed it!

PikachuEXE:
OK I get this is a proposal now

I would ask
> 1. Is there anything would be taken away (including the potential benefits) by moving away from `prepend`?
> 2. How this gem's compatibility with other gems (which also use meta programming, e.g. [contracts.ruby](https://github.com/egonSchiele/contracts.ruby)) would be comparing `prepend` and `extend`?
> 3. Is there any other not so obvious potential issues that we can think of?
> 4. What are the original intents by picking `prepend` over `extend`/`include`?

paracycle:
@PikachuEXE all of those points will be addressed in the PR description in detail once the build gets green on this branch.

PikachuEXE:
> This branch has conflicts that must be resolved

paracycle:
> > This branch has conflicts that must be resolved
> 
> 

Yes! That's why the PR is marked [WIP] and is set as Draft.

jemmaissroff:
Closed and replaced by #253

PR #250, opened 2021-12-13
Revert internal data structure to simple hash

### LINKED ISSUE #251: Array data structure leading to multiple bugs
@PikachuEXE noted in [this comment](https://github.com/panorama-ed/memo_wise/issues/238#issuecomment-992134835):

> Now I am getting random `memo_wise`ed methods returning unexpected values on each request (e.g. `nil` returned for methods with objects, method expected one class returning object of another class)
>
> Not sure how to reproduce with minimal example yet
>
> Update 1: Sometimes it returns value like `{[[], {}]=>nil}` which seems to be what's used in caching

In our own codebase we're seeing MemoWise v1.4 yielding this type of error as well. It appears to be yet another case of multiple methods sharing the same `index` on the same object, despite attempts to fix this type of issue in both v1.3 and v1.4. What @PikachuEXE describes above is a zero-argument method returning the data structure for a multi-argument method, but we also see it in reverse, with methods that take arguments getting `true` in the array and attempting to use it as their intermediate hash, leading to ``undefined method `[]' for true:TrueClass``.

In addition, we've noticed at least one _other_ (🙀) type of error: `SystemStackError: stack level too deep` occurring when somehow the `klass.send(method_name)` does not invoke the method defined with `klass.module_eval` and so the `send` recursively calls itself forever.

JacobEvelyn:
Previous versions of `MemoWise` used an array as the
internal cache data structure, where each memoized method
uses a unique array index. However, complex inheritance
structures make it easy to have multiple methods accidentally
share the same array index on the same object, resulting
in nondeterministic and incorrect behavior.

While we may explore that optimization again in the future,
for the time being we are switching to a hash data structure
keyed on method name, which is safer because two methods
cannot share the same name in Ruby.

Fixes #251 

**Before merging:**

### DIFF
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -76,14 +76,6 @@ BENCHMARK_GEMS.each do |benchmark_gem|
       end
       #{benchmark_gem.memoization_method} :no_args
 
-      # For the no_args case, we can't depend on arguments to alternate between
-      # returning truthy and falsey values, so instead make two separate
-      # no_args methods
-      def no_args_falsey
-        nil
-      end
-      #{benchmark_gem.memoization_method} :no_args_falsey
-
       def one_positional_arg(a)
         100 if a.positive?
       end
@@ -132,9 +124,7 @@ end
 #
 # NOTE: The proportion of falsey results is 1/N_UNIQUE_ARGUMENTS (because for
 # the methods with arguments we are truthy for all but the first unique argument
-# set, and for zero-arity methods we manually execute `no_args` N_TRUTHY_RESULTS
-# times per each execution of `no_args_falsey`). This number was selected as the
-# lowest number such that this logic:
+# set). This number was selected as the lowest number such that this logic:
 #
 #   output = hash[key]
 #   if output || hash.key?(key)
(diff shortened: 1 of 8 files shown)


jemmaissroff (approved):

benchmarks/benchmarks.rb
```diff
@@ -132,9 +124,7 @@ def positional_splat_keyword_and_double_splat_args(a, *args, b:, **kwargs)
 #
 # NOTE: The proportion of falsey results is 1/N_UNIQUE_ARGUMENTS (because for
 # the methods with arguments we are truthy for all but the first unique argument
-# set, and for zero-arity methods we manually execute `no_args` N_TRUTHY_RESULTS
-# times per each execution of `no_args_falsey`). This number was selected as the
-# lowest number such that this logic:
```
ms-ati, line 137: @JacobEvelyn why remove the falsey tests from the benchmarks -- just not clear if this is on purpose?

ms-ati (approved):
LGTM

JacobEvelyn:
Thank you @PikachuEXE !

PikachuEXE:
I just notice the branch name is `back-to-the-future`... 🤣

PR #255, opened 2021-12-17
Fix race conditions in methods with arguments

JacobEvelyn:
This commit fixes race conditions found in memoized methods
that take arguments. This comes at a slight performance cost
for truthy memoized values (due to switching from `Hash#[]`
to the slightly slower `Hash#fetch`) but will be faster when
memoized values are falsey.

This change removes the "truthy optimization" leading to these
race conditions and as a result allows us to simplify our
benchmark code as we no longer need to benchmark truthy and
falsey values separately.

**Before merging:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@
   * Support for memoization on frozen objects
   * Support for memoization of class and module methods
   * Support for inheritance of memoized class and instance methods
+  * Documented and tested [thread-safety guarantees](#thread-safety)
   * Full [documentation](https://rubydoc.info/github/panorama-ed/memo_wise/MemoWise) and [test coverage](https://codecov.io/gh/panorama-ed/memo_wise)!
 
 ## Installation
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -17,6 +17,7 @@
 `MemoWise` is **the wise choice for Ruby memoization**, featuring:
 
   * Fast performance of memoized reads (with [benchmarks](#benchmarks))
+  * Strong [thread-safety guarantees](#thread-safety)
```
JacobEvelyn, line 20: Open to changing the order here!

ms-ati: Might we consider "Documented and tested thread-safety guarantees" as an alternative to "Strong"? In thread-safety terminology, I think "Strong" may imply to some people that we have our own mutexes for example

JacobEvelyn: Yeah I wondered if "strong" would have specific meanings here. Will change!

ms-ati: I also might have placed this at the bottom (under the API aspects), as I don't believe thread-safety guarantees being spelled out and tested is the very 2nd thing most are looking for?

ms-ati (approved):
LGTM

PR #265, opened 2022-01-07
Fix memoization with inheritance

### LINKED ISSUE #262: Cannot use memoized class level method in sub classes
```ruby

require "memo_wise"

module M1
  module Concern
    def self.included(includer)
      includer.extend ClassMethods
    end
  end

  module ClassMethods
    prepend MemoWise

    # This rarely changed
    # So should be fine caching on first usage
    def class_level_method
      1 + 1
    end
    memo_wise :class_level_method
  end
end

class C1
  include M1::Concern

  def instance_level_method
    self.class.class_level_method + 1
  end
end

class C2 < C1
  # Empty
end

# These are fine
C1.class_level_method
C1.new.instance_level_method

# These are not
C2.class_level_method
C2.new.instance_level_method
```

I use this pattern in my controllers
C1 would be `ApplicationController` and C2 would be my any of the used controllers
M1 would be any of the app-wise shared controller mixin modules

jemmaissroff:
Fixes #262 

**Before merging:**

### DIFF
--- a/spec/memo_wise_spec.rb
+++ b/spec/memo_wise_spec.rb
@@ -630,6 +630,43 @@ RSpec.describe MemoWise do
           end
         end
       end
+
+      context "when a class inherits from a parent class whose extended module memoizes methods" do
+        let(:parent_class) do
+          Class.new do
+            extend Module1
+          end
+        end
+
+        let(:module1) do
+          Module.new do
+            prepend MemoWise
+
+            def module1_method_counter
+              @module1_method_counter || 0 # rubocop:disable RSpec/InstanceVariable
+            end
+
+            def module1_method
+              @module1_method_counter = module1_method_counter + 1
+              "module1_method"
+            end
+            memo_wise :module1_method
+          end
+        end
+
+        let(:child_class) do
+          Class.new(parent_class)
+        end
+
+        before(:each) do
+          stub_const("Module1", module1)
+        end
+
+        it "memoizes inherited methods separately" do
+          expect(Array.new(4) { child_class.module1_method }).to all eq("module1_method")
+          expect(child_class.module1_method_counter).to eq(1)
+        end
+      end
     end
 
     context "with module mixed into other classes" do
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -156,16 +156,14 @@ module MemoWise
           klass = klass.singleton_class
         end
 
-        if klass.singleton_class?
-          # This ensures that a memoized method defined on a parent class can
-          # still be used in a child class.
-          klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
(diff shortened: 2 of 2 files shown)


spec/inheritance_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+          "class_method"
+        end
+
+        def instance_method
+          "instance_method"
+        end
+      end
+    end
+
+    it { expect(subject.class_method).to eq("class_method") }
```
jemmaissroff, line 21: These two tests were unrelated to the specific issue, just seemed helpful to include

spec/inheritance_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      memo_wise :class_method
+    end
+
+    let(:inherited_klass) do
+      Class.new do
+        extend module_with_class_method
+      end
+    end
+
+    it { expect(subject.class_method).to eq("class_method") }
```
jemmaissroff, line 41: Confirmed that this test was previously failing

spec/inheritance_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+RSpec.describe "inheritance" do # rubocop:disable RSpec/DescribeClass
```
JacobEvelyn, line 3: I don't feel too strongly but I have a preference for keeping these tests in the `spec/memo_wise_spec.rb` file because we have a whole host of other inheritance examples in there and I find it easier to mentally sort out what cases we've tested (for example, this isn't just "simple inheritance," it's also class methods defined with `self.` as opposed to the other types of methods we support).

Thoughts?

jemmaissroff: I think it's more clear in a separate file (and could even migrate the existing ones here). The existing ones are inserted as parts of other contexts instead of all in one place, and I think it's more straightforward to keep inheritance as its own set of specs. Let me know your thoughts

spec/memo_wise_spec.rb
```diff
@@ -630,6 +630,45 @@ def self.module_method; end
...
+            end
+            memo_wise :module1_method
+          end
+        end
+
+        let(:child_class) do
+          Class.new(parent_class)
+        end
+
+        let(:instance) { child_class.new }
```
JacobEvelyn, line 661: Do we need `instance` at all?

JacobEvelyn (approved):
LGTM if:

1. you remove `instance` (or correct me if I'm missing something).
2. you confirm that the test fails without your fix (which I think it does but just want to make sure since it's been a while).

jemmaissroff:
> 1. you remove `instance` (or correct me if I'm missing something).
Done!

> 2. you confirm that the test fails without your fix (which I think it does but just want to make sure since it's been a while).
Confirmed!

PR #271, opened 2022-02-28
Fix Github Actions CI for Ruby 3.0

ms-ati:
Due to an open issue with Github Actions [1], currently
the version 3.0 is parsed as 3, and leads to the latest
Ruby version being used which is currently 3.1.1.

The fix is to use quotes, e.g. '3.0'.

[1]: https://github.com/actions/runner/issues/849


.github/workflows/main.yml
```diff
@@ -12,7 +12,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [jruby, 2.4, 2.5, 2.6, 2.7, 3.0, 3.1, truffleruby-head]
+        # Due to https://github.com/actions/runner/issues/849, we have to use
+        # quotes for '3.0' -- without quotes, CI sees '3' and runs the latest
+        # which is currently 3.1.1 as of 2022-02-28.
```
jemmaissroff, line 17: Very nitpicky, but do we need this line `which is currently 3.1.1 as of 2022-02-28.`?

jemmaissroff (approved):

.github/workflows/main.yml
ms-ati: Sure, I'm happy to remove that

ms-ati: Updated!

PR #274, opened 2022-03-16
Nuanced benchmarking

jemmaissroff:
This PR allows us to more easily benchmark against the version of MemoWise on the main branch in GitHub. In the GitHub actions to print benchmarks, we now also print benchmarks of running the MemoWise main branch against whatever branch a PR creates. This creates an easier way to see if a change is actually more performant for both local development and PR review.

**Before merging:**

### DIFF
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -3,7 +3,26 @@
 require "benchmark/ips"
 
 require "tempfile"
-require "memo_wise"
+
+github_memo_wise_path = Gem.loaded_specs["memo_wise"].full_gem_path
+
+# This string is both used for temp filepaths necessary to separate the GitHub
+# version of MemoWise and the local version, and used for the reported results
+GITHUB_MAIN = "MemoWise_GitHubMain"
+
+# We download a the main branch of MemoWise on GitHub into a tmp directory to
+# compare against the local version when we run benchmarks
+Dir.mktmpdir do |directory|
+  Dir["#{github_memo_wise_path}/lib/**/*.rb"].each do |file|
+    Tempfile.open([File.basename(file)[0..-4], ".rb"], directory) do |tempfile|
+      tempfile.write(File.read(file).gsub("MemoWise", GITHUB_MAIN))
+      tempfile.rewind
+      require tempfile.path
+    end
+  end
+end
+
+require_relative "../lib/memo_wise"
 
 # Some gems do not yet work in Ruby 3 so we only require them if they're loaded
 # in the Gemfile.
@@ -51,6 +70,7 @@ end
 # NOTE: Some gems do not yet work in Ruby 3 so we only test with them if they've
 # been `require`d.
 BENCHMARK_GEMS = [
+  BenchmarkGem.new(MemoWise_GitHubMain, "prepend #{GITHUB_MAIN}", :memo_wise),
   BenchmarkGem.new(MemoWise, "prepend MemoWise", :memo_wise),
   (BenchmarkGem.new(DDMemoize, "DDMemoize.activate(self)", :memoize) if defined?(DDMemoize)),
   (BenchmarkGem.new(Dry::Core, "include Dry::Core::Memoizable", :memoize) if defined?(Dry::Core)),
@@ -187,7 +207,7 @@ benchmark_lambdas = [
 
 # We benchmark different cases separately, to ensure that slow performance in
 # one method or code path isn't hidden by fast performance in another.
-benchmark_lambdas.map do |benchmark|
+benchmark_jsons = benchmark_lambdas.map do |benchmark|
   json_file = Tempfile.new
 
   Benchmark.ips do |x|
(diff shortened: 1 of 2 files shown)


ms-ati:
Cool! Are we to copy over the results including the GithubMain comparison? Do you want to include an example output or should we run this branch ourselves to see it?

jemmaissroff:
> Cool! Are we to copy over the results including the GithubMain comparison? Do you want to include an example output or should we run this branch ourselves to see it?

I should've said this on the PR itself, but the output is already [in the benchmarks run from this PR](https://github.com/panorama-ed/memo_wise/runs/5571470286?check_suite_focus=true). Copying it over here too if it's easier to see:

```
|Method arguments|`GitHubMain` (1.6.0)|
|--|--|
|`()` (none)|0.99x|
|`(a)`|0.99x|
|`(a, b)`|0.97x|
|`(a:)`|0.99x|
|`(a:, b:)`|0.98x|
|`(a, b:)`|0.97x|
|`(a, *args)`|0.98x|
|`(a:, **kwargs)`|0.97x|
|`(a, *args, b:, **kwargs)`|0.99x|

|Method arguments|`Dry::Core` (0.7.1)|`Memery` (1.4.0)|
|--|--|--|
|`()` (none)|1.07x|12.17x|
|`(a)`|1.63x|9.80x|
|`(a, b)`|0.33x|1.98x|
|`(a:)`|1.53x|18.41x|
|`(a:, b:)`|0.36x|4.22x|
|`(a, b:)`|0.35x|4.13x|
|`(a, *args)`|0.81x|1.91x|
|`(a:, **kwargs)`|0.78x|3.19x|
|`(a, *args, b:, **kwargs)`|0.57x|1.51x|
```

benchmarks/benchmarks.rb
```diff
@@ -3,18 +3,33 @@
 require "benchmark/ips"
 
 require "tempfile"
-require "memo_wise"
+
+github_memo_wise_path = Gem.loaded_specs["memo_wise"].full_gem_path
+
+GITHUB_MAIN = "GitHubMain"
```
ms-ati, line 9: MINOR: In the results, I was wondering if this might be clearer as "MemoWise-GitHubMain" or otherwise indicating that it's the Github main branch of MemoWise?

ms-ati: MINOR: Can we have a comment explaining that this string will be used both for the temp paths, as well as the reported results, for the version of MemoWise currently in Github main branch?

benchmarks/benchmarks.rb
```diff
@@ -3,18 +3,33 @@
...
+
+Dir.mktmpdir do |directory|
```
ms-ati, line 11: MINOR: Can we have a comment explaining just that we are going to download the version of MemoWise from Github main branch into a temp directory here, to compare benchmarks against?

benchmarks/benchmarks.rb
```diff
@@ -3,18 +3,33 @@
...
+
+require_relative "../lib/memo_wise"
 
 # Some gems do not yet work in Ruby 3 so we only require them if they're loaded
 # in the Gemfile.
 %w[memery memoist memoized memoizer ddmemoize dry-core].
   each { |gem| require gem if Gem.loaded_specs.key?(gem) }
 
-# The VERSION constant does not get loaded above for these gems.
+# # The VERSION constant does not get loaded above for these gems.
```
ms-ati, line 28:
```suggestion
# The VERSION constant does not get loaded above for these gems.
```

benchmarks/benchmarks.rb
```diff
@@ -3,18 +3,33 @@
...
 %w[memoized memoizer].
   each { |gem| require "#{gem}/version" if Gem.loaded_specs.key?(gem) }
 
-# The Memoizable module from dry-core needs to be required manually
+# # The Memoizable module from dry-core needs to be required manually
```
ms-ati, line 32:
```suggestion
# The Memoizable module from dry-core needs to be required manually
```

ms-ati (approved):
Hi Jemma this all looks good to me -- I left a couple MINOR requests for comments to clarify, and perhaps a renaming of "GitHubMain" to make the results a bit clearer to a new reader?

benchmarks/benchmarks.rb
```diff
@@ -203,42 +223,50 @@ def positional_splat_keyword_and_double_splat_args(a, *args, b:, **kwargs)
...
-    "%.#{N_RESULT_DECIMAL_DIGITS}fx" %
-      (memo_wise["central_tendency"] / bgem["central_tendency"]).round(N_RESULT_DECIMAL_DIGITS)
-  end.join("|")
-
-  name = memo_wise["name"].partition(": ").last
-  puts "|`#{name}`#{' (none)' if name == '()'}|#{output_str}|"
+    name = memo_wise["name"].partition(": ").last
+    puts "|`#{name}`#{' (none)' if name == '()'}|#{output_str}|"
+  end
+  puts ""
```
ms-ati, line 271: Is this last "puts" on purpose?

jemmaissroff: Yep, it allows for space between the two outputs. Definitely let me know if you can think of a clearer way to do this

ms-ati: Ah ok, maybe just a comment making the intent clear - I'll leave a suggestion

PR #276, opened 2022-04-01
Use nested hashes for multiple required arguments

### LINKED ISSUE #244: Use nested hashes instead of arrays for methods with multiple required arguments
This was a suggestion made by @brasic in the RubyConf Discord channel for our talk. The idea is that for a method like `foo(a, b)`, instead of storing the results like this:

```ruby
# @_memo_wise[index_of_foo]
{ [a, b] => memoized_value }
```

we'd store them like this:

```ruby
# @_memo_wise[index_of_foo]
{ a => { b => memoized_value } }
```

our talk implies that this would be faster (hash lookups on array keys are significantly slower, and in the already-memoized case we avoid a new array allocation each time), and in Discord @brasic had a microbenchmark showing this solution to be faster as well.

JacobEvelyn:
This commit optimizes methods with multiple required arguments to use
nested hashes to store the memoized value (so each argument accesses
the next layer in the hash) rather than a single-layer hash with an
array key.

This is more performant because multiple `#hash` calls on simpler
data structures are faster than one `#hash` lookup on all of those
elements within an array, and in addition we avoid the cost of
initializing an array on every lookup.

Closes #244 

**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -195,9 +195,40 @@ module MemoWise
               end
             end
           HEREDOC
-        # MemoWise::InternalAPI::MULTIPLE_REQUIRED, MemoWise::InternalAPI::SPLAT,
-        # MemoWise::InternalAPI::DOUBLE_SPLAT, MemoWise::InternalAPI::SPLAT_AND_DOUBLE_SPLAT
-        else
+        when MemoWise::InternalAPI::MULTIPLE_REQUIRED
+          # When we have multiple required params, we store the memoized values in a deeply nested hash, like:
+          # { method_name: { arg1 => { arg2 => { arg3 => memoized_value } } } }
+          last_index = method.parameters.size
+          layers = method.parameters.map.with_index(1) do |(_, name), index|
+            prev_hash = "_memo_wise_hash#{index - 1 if index > 1}"
+            fallback = if index == last_index
+                         "#{original_memo_wised_name}(#{MemoWise::InternalAPI.call_str(method)})"
+                       else
+                         "{}"
+                       end
+            "_memo_wise_hash#{index} = #{prev_hash}.fetch(#{name}) { #{prev_hash}[#{name}] = #{fallback} }"
+          end
+          klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
+            def #{method_name}(#{MemoWise::InternalAPI.args_str(method)})
+              _memo_wise_hash = (@_memo_wise[:#{method_name}] ||= {})
+              #{layers.join("\n  ")}
+            end
+          HEREDOC
+        when MemoWise::InternalAPI::SPLAT_AND_DOUBLE_SPLAT
+          # When we have both *args and **kwargs, we store the memoized values in a deeply nested hash, like:
+          # { method_name: { args => { kwargs => memoized_value } } }
+          klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
+            def #{method_name}(*args, **kwargs)
+              _memo_wise_hash = (@_memo_wise[:#{method_name}] ||= {})
+              _memo_wise_kwargs_hash = _memo_wise_hash.fetch(args) do
+                _memo_wise_hash[args] = {}
+              end
+              _memo_wise_kwargs_hash.fetch(kwargs) do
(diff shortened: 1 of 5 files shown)


JacobEvelyn:
From this [build run](https://github.com/panorama-ed/memo_wise/runs/5791188790?check_suite_focus=true) with Ruby 3.1.1:

|Method arguments|`MemoWise_GitHubMain` (1.6.0)|
|--|--|
|`()` (none)|1.06x|
|`(a)`|1.06x|
|`(a, b)`|3.54x|
|`(a:)`|1.01x|
|`(a:, b:)`|3.10x|
|`(a, b:)`|3.19x|
|`(a, *args)`|0.98x|
|`(a:, **kwargs)`|1.00x|
|`(a, *args, b:, **kwargs)`|1.33x|

spec/reset_memo_wise_spec.rb
```diff
@@ -267,8 +267,9 @@
             expect { target.reset_memo_wise(:with_optional_positional_and_keyword_args) }.not_to raise_error
           end
 
-          it "doesn't raise an error for methods with optional positional and keyword arguments provided" do
-            expect { target.reset_memo_wise(:with_optional_positional_and_keyword_args, a: 1) }.not_to raise_error
+          it "doesn't raise an error for methods with optional positional and keyword arguments partially provided" do
```
JacobEvelyn, line 270: Changing this test keeps us at 100% test coverage.

lib/memo_wise.rb
```diff
@@ -195,9 +195,39 @@ def #{method_name}(#{MemoWise::InternalAPI.args_str(method)})
...
+              _memo_wise_hash = (@_memo_wise[:#{method_name}] ||= {})
+              _memo_wise_kwargs_hash = _memo_wise_hash.fetch(args) do
+                _memo_wise_hash[args] = {}
+              end
+              _memo_wise_kwargs_hash.fetch(kwargs) do
+                _memo_wise_kwargs_hash[kwargs] = #{original_memo_wised_name}(#{MemoWise::InternalAPI.call_str(method)})
+              end
+            end
+          HEREDOC
+        else # MemoWise::InternalAPI::SPLAT, MemoWise::InternalAPI::DOUBLE_SPLAT
```
ms-ati, line 230: Didn't there used to be `MemoWise::InternalAPI::SPLAT_AND_DOUBLE_SPLAT` as well?

ms-ati: Oops, didn't see it! nevermind

lib/memo_wise.rb
```diff
@@ -530,15 +568,22 @@ def reset_memo_wise(method_name = nil, *args, **kwargs)
...
-      key = if method_arguments == MemoWise::InternalAPI::SPLAT_AND_DOUBLE_SPLAT
-              [args, kwargs]
-            else
-              method.parameters.map.with_index do |(type, name), i|
-                type == :req ? args[i] : kwargs[name]
-              end
-            end
-      method_hash&.delete(key)
+    when MemoWise::InternalAPI::SPLAT_AND_DOUBLE_SPLAT
+      method_hash&.[](args)&.delete(kwargs)
```
ms-ati, line 572: I'm confused why the key for `SPLAT_AND_DOUBLE_SPLAT` doesn't continue to be `[args, kwargs]`? Is it because the `args` is handled already somehow?

JacobEvelyn: The premise of this PR is that for any values, it is faster to do:

```ruby
hash[val1][val2]
```

than it is to do:

```ruby
hash[[val1, val2]]
```

because of both simpler/faster `.hash` methods, and avoiding the instantiation of a temporary array. This premise is true even when `val1` and `val2` are `args` and `kwargs`. (This is why we see benchmark speedups for the `(a, *args, b:, **kwargs)` case.)

Does that make sense?

ms-ati: Yes! Thank you!

But could we have a comment on that case, saying “first time we lookup with `args`, next time through we’ll be nested on level down and end up at (which branch so we end up at to do the `kwargs`?)

JacobEvelyn: Just added some clarifying comments. Let me know if it's still confusing.

ms-ati (approved):
Amazing!

lib/memo_wise.rb
ms-ati: Thank you!

PR #278, opened 2022-04-04
Update CHANGELOG.md for 1.7.0

JacobEvelyn:
**Before merging:**

- [ ] ~Copy the table printed at the end of the latest benchmark results into the `README.md` and update this PR~
- [ ] ~If this change merits an update to `CHANGELOG.md`, add an entry following Keep a Changelog [guidelines](https://keepachangelog.com/en/1.0.0/) with [semantic versioning](https://semver.org/)~

PR #248, opened 2021-12-10
Update CHANGELOG.md for v1.4.0

JacobEvelyn:
**Before merging:**

- [x] ~Copy the table printed at the end of the latest benchmark results into the `README.md` and update this PR~
- [x] If this change merits an update to `CHANGELOG.md`, add an entry following Keep a Changelog [guidelines](https://keepachangelog.com/en/1.0.0/) with [semantic versioning](https://semver.org/)

PR #282, opened 2022-04-11
Restructure CHANGELOG.md for clarity and consistency

### LINKED ISSUE #242: Tidy CHANGELOG
Our CHANGELOG has some inconsistent formatting that could be improved. For example:

- Tenses: do we use `fix` or `fixed` or `fixes`?
- Subheadings: should all versions use the same subheadings?
- PR links: can we link to associated PRs as mentioned in https://github.com/panorama-ed/memo_wise/pull/241#pullrequestreview-824741345 ?
- Diff links: is it clearer to put diff links inline instead of at the bottom of the doc? (I recently released a new version and didn't realize I needed to add that so the link was broken.)

JacobEvelyn:
Closes #242

**Some things I would love reviewer input on:**

- Overall structure. I thought "Gem enhancements" and "Project enhancements" made sense but I'm open to any and all feedback.
- Tenses. I went back and forth so many times on "added" vs. "add" vs. "has been added" etc.
- Breaking changes. I think it's helpful to show them, but realistically we only have one entry so I wasn't sure if there was a better/more minimal way to list them (I tried putting it within the `Gem enhancements:` section and a few other alternatives and don't love any of them).
- PR links. They're nice but I didn't want to go back and add them all so I just removed the few we had. 🤷‍♀️ 

**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,156 +1,232 @@
 # Changelog
 
-All notable changes to this project will be documented in this file.
+All notable changes to this project will be documented in this file, which
+follows a format inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [Unreleased](https://github.com/panorama-ed/memo_wise/compare/v1.7.0...HEAD)
 
-- Nothing yet!
+**Gem enhancements:** none
 
-## [1.7.0] - 2022-04-04
+_No breaking changes!_
 
-### Updated
+**Project enhancements:** none
 
-- Optimize memoized lookups for methods with multiple required arguments
-  ([#276](https://github.com/panorama-ed/memo_wise/pull/276))
+## [v1.7.0](https://github.com/panorama-ed/memo_wise/compare/v1.6.0...v1.7.0) - 2022-04-04
 
-## [1.6.0] - 2022-01-24
+**Gem enhancements:**
 
-### Fixed
+- Optimized memoized lookups for methods with multiple required arguments
+  [[#276](https://github.com/panorama-ed/memo_wise/pull/276)]
 
-- Fixed a bug relating to inheritance of classes which include module which
-  prepends MemoWise ([#265](https://github.com/panorama-ed/memo_wise/pull/265))
+_No breaking changes!_
 
-### Updated
+**Project enhancements:**
 
-- Update official test coverage to support Ruby 3.1
+- Added benchmarking against GitHub `main` branch to CI [[#274](https://github.com/panorama-ed/memo_wise/pull/274)]
 
-## [1.5.0] - 2021-12-17
+## [v1.6.0](https://github.com/panorama-ed/memo_wise/compare/v1.5.0...v1.6.0) - 2022-01-24
 
-### Fixed
+**Gem enhancements:**
 
-- Remove optimization for truthy results to fix thread safety race condition
-  bugs
-- Switch to a simpler internal data structure to fix several classes of bugs
-  related to inheritance that the previous few versions were unable to
-  sufficiently address
+- Fixed a bug relating to inheritance of classes which include a module which
+  `prepend`s `MemoWise` [[#262](https://github.com/panorama-ed/memo_wise/pull/262)]
 
-## [1.4.0] - 2021-12-10
+_No breaking changes!_
 
-### Fixed
+**Project enhancements:**
 
-- Fix several bugs related to classes inheriting memoized methods
-  from multiple modules or a parent class ([#241](https://github.com/panorama-ed/memo_wise/pull/241))
+- Update official test coverage to support Ruby 3.1 [[#263](https://github.com/panorama-ed/memo_wise/pull/263)]
 
-## [1.3.0] - 2021-11-22
+## [v1.5.0](https://github.com/panorama-ed/memo_wise/compare/v1.4.0...v1.5.0) - 2021-12-17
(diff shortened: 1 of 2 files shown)


ms-ati:
Q: Do we see value in trying to stay closely in line with the documented format of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)?

JacobEvelyn:
> Q: Do we see value in trying to stay closely in line with the documented format of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)?

Oh! I hadn't realized that Keep a Changelog was prescriptive about the sub-headings.

I'm open to pushback, but there's a few things I think are not ideal about Keep a Changelog's format:

1. Having the links at the bottom makes them prone to copy-pasta typos. I've made errors there [recently](https://github.com/panorama-ed/memo_wise/pull/278#commitcomment-70595708) and [previously](https://github.com/panorama-ed/memo_wise/pull/248#discussion_r766778460) a [few times](https://github.com/panorama-ed/memo_wise/pull/223/files). It feels easier to me to have those links inline.
2. The big things I usually want to get out of a changelog are "what changed?" and "do I need to change my code when I upgrade?" I get that semver is supposed to address the second question, but I find it's rarely used consistently/correctly, including in this project (e.g. many of our changes should be patch releases instead of minor releases). And even when semver is used, you still want to know what you need to change when you upgrade, so I think clearly specifying breaking changes is helpful.
3. Our changelog has historically put both user-relevant changes (e.g. performance optimizations and bugfixes) alongside user-irrelevant changes (e.g. linter and benchmark adjustments). I think it's much clearer to separate these.

Would love to hear others' thoughts.

ms-ati:
Note: I'm totally sold on moving to in-line links instead of footnote-style links, to avoid  the errors

ms-ati:
What if we:

- kept the KeepAChangelog "Added, Updated, Fixed"
- added a **new header** for "Internal enhancements"
- Made "breaking change!" a sub-bullet that jumps out on any item, rather than a section?

JacobEvelyn:
> What if we:
> 
> * Made "breaking change!" a sub-bullet that jumps out on any item, rather than a section?

I considered this, but I don't love that it doesn't tell you of the _absence_ of any breaking changes. Thoughts?

JacobEvelyn:
> One note here though - it confuses me why we include the breaking changes section with "None!" when there's nothing, but don't do the same for "Project Enhancements"

@jemmaissroff what sort of breaking "project enhancements" are you thinking about? From my perspective breaking changes are really for our users, not our (very very few) developers, and I don't consider a linter or benchmark change to be "breaking" in the same way that I do a code change.

I actually think I might further push breaking changes into the code enhancements section, like:

```markdown
### Gem enhancements:

- Fixed a bug relating to inheritance of classes which include a module which
  `prepend`s `MemoWise`

*Breaking changes: none*

### Project enhancements:

- Blah blah blah
```

jemmaissroff:
> @jemmaissroff what sort of breaking "project enhancements" are you thinking about? From my perspective breaking changes are really for our users, not our (very very few) developers, and I don't consider a linter or benchmark change to be "breaking" in the same way that I do a code change.

I didn't explain this well at all, that's my fault. What I mean is, it confuses me that if we have no "breaking changes" we still keep the section for it (and write "None!"), while if we have no "project enhancements" we don't keep the section for it. Does that make more sense?

JacobEvelyn:
> What I mean is, it confuses me that if we have no "breaking changes" we still keep the section for it (and write "None!"), while if we have no "project enhancements" we don't keep the section for it. Does that make more sense?

Yes that makes sense! Would any of these options feel better to you?

```markdown
### Code enhancements (no breaking changes):

- a
- b

### Project enhancements: none
```

```markdown
**Code enhancements (no breaking changes):**

- a
- b

**Project enhancements: none**
```

```markdown
**Code enhancements:**

- a
- b

_Breaking changes: none_

**Project enhancements: none**
```

```markdown
### Code enhancements:

- a
- b

_Breaking changes: none_

### Project enhancements: none
```

```markdown
### Code enhancements:

- a
- b

_Breaking changes: none_

### Project enhancements:

- None
```

I think I prefer options 2 or 3. The headers feel a bit overwhelming to me, especially when no real content follows.

jemmaissroff:
Yup, option 3 makes a lot of sense to me too!

CHANGELOG.md
```diff
@@ -2,155 +2,188 @@
 
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
JacobEvelyn, line 5: Should I delete this too?

ms-ati: I think we still aim to follow Semantic Versioning -- perhaps I am out of the loop? The fact it wasn't always followed (especially on the way to v1.0.0!) I don't see as changing that?

CHANGELOG.md
```diff
@@ -2,155 +2,188 @@
 
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
```
ms-ati, line 5: It appears to me that we are staying extremely close to the Keep A Changelog format, and it seems to me we are still following all the most important conventions (e.g. an "Unreleased" section at the top), but only changing two details (what headers under each release, and inline vs footnote-style links) -- does that match your understanding?

If so, would we consider keeping the link and saying something like:

```
The format is inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), but has minor changes to the headers we use to breakout changes in each release.
```

ms-ati:
@JacobEvelyn I'm fine with all he format changes, but I hope we can align on (a) KeepAChangelog and (b) Semantic Versioning, especially the latter?

ms-ati (approved):

Issue #302, opened 2023-03-15
memo_wise raises error when memoizing a method in a module that's included in a class that overrides `initialize`


joevandyk:
I might be using memo_wise incorrectly here.. but seeing some odd behavior when I memoize a method in a module that's included in a class that has an `initialize` method.

```
require 'memo_wise'

module ModuleMethods
  prepend MemoWise

  def a_module_method
    self.class.name
  end
  memo_wise :a_module_method
end

# Works to call ClassWithoutInitializer.new.a_module_method
class ClassWithoutInitializer
  include ModuleMethods
end

# Does not work to call ClassWithInitializer.new("thing").a_module_method
class ClassWithInitializer
  include ModuleMethods

  def initialize(name)
    @name = name
  end
end

# Works
puts ClassWithoutInitializer.new.a_module_method

# Does not Work
# memo_wise-1.7.0/lib/memo_wise.rb:183:in `a_module_method': undefined method `fetch' for nil:NilClass (NoMethodError)
puts ClassWithInitializer.new("name").a_module_method
```

joevandyk:
https://github.com/panorama-ed/memo_wise/pull/321 seems to fix the issue for us.

dburgess560:
Within the example, adding a `prepend MemoWise` sorts out the memoization cache instantiation issue. Not amazing but bypasses the issue for now.
```
class ClassWithInitializer
  prepend MemoWise
  include ModuleMethods

  def initialize(name)
    @name = name
  end
end
```

PR #321, opened 2023-10-31
always set memo_wise hash if not set for some reason

joevandyk:
Fixes https://github.com/panorama-ed/memo_wise/issues/302 for us

**Before merging:**

### DIFF
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -181,6 +181,7 @@ module MemoWise
         when MemoWise::InternalAPI::NONE
           klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
             def #{method_name}
+              @_memo_wise ||= {}
               @_memo_wise.fetch(:#{method_name}) do
                 @_memo_wise[:#{method_name}] = #{original_memo_wised_name}
               end
@@ -190,6 +191,7 @@ module MemoWise
           key = method.parameters.first.last
           klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
             def #{method_name}(#{MemoWise::InternalAPI.args_str(method)})
+              @_memo_wise ||= {}
               _memo_wise_hash = (@_memo_wise[:#{method_name}] ||= {})
               _memo_wise_hash.fetch(#{key}) do
                 _memo_wise_hash[#{key}] = #{original_memo_wised_name}(#{MemoWise::InternalAPI.call_str(method)})
@@ -211,6 +213,7 @@ module MemoWise
           end
           klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
             def #{method_name}(#{MemoWise::InternalAPI.args_str(method)})
+              @_memo_wise ||= {}
               _memo_wise_hash = (@_memo_wise[:#{method_name}] ||= {})
               #{layers.join("\n  ")}
             end
@@ -220,6 +223,7 @@ module MemoWise
           # { method_name: { args => { kwargs => memoized_value } } }
           klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
             def #{method_name}(*args, **kwargs)
+              @_memo_wise ||= {}
               _memo_wise_hash = (@_memo_wise[:#{method_name}] ||= {})
               _memo_wise_kwargs_hash = _memo_wise_hash.fetch(args) do
                 _memo_wise_hash[args] = {}
@@ -232,6 +236,7 @@ module MemoWise
         else # MemoWise::InternalAPI::SPLAT, MemoWise::InternalAPI::DOUBLE_SPLAT
           klass.module_eval <<~HEREDOC, __FILE__, __LINE__ + 1
             def #{method_name}(#{MemoWise::InternalAPI.args_str(method)})
+              @_memo_wise ||= {}
               _memo_wise_hash = (@_memo_wise[:#{method_name}] ||= {})
               _memo_wise_key = #{MemoWise::InternalAPI.key_str(method)}
               _memo_wise_hash.fetch(_memo_wise_key) do
@@ -552,6 +557,7 @@ module MemoWise
       raise ArgumentError, "Provided args when method_name = nil" unless args.empty?
       raise ArgumentError, "Provided kwargs when method_name = nil" unless kwargs.empty?
 
+      @_memo_wise ||= {}
       @_memo_wise.clear
       return
     end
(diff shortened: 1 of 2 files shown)


JacobEvelyn:
Interesting, thanks @joevandyk ! Would you be able to write a failing test that passes with this change? If you need help navigating our test files let me know and I can try to add one based on your previous example.

joevandyk:
I'm at a bit of a loss. I updated the PR here with some sample tests and a failing test script. 

I can't get the rspec tests to fail. 

I added `spec/failing.rb`. It only seems to fail if I don't use `bundle exec`. 

I do see the error if I run `bundle exec irb` from inside the memo_wise repo and copy/paste the module and class and test code from the github issue into IRB. 

```
➜ $ bundle exec ruby spec/failing.rb
ClassWithoutInitializer
ClassWithInitializer

➜  $ ruby spec/failing.rb
ClassWithoutInitializer
/Users/joe/.asdf/installs/ruby/3.2.1/lib/ruby/gems/3.2.0/gems/memo_wise-1.8.0/lib/memo_wise.rb:184:in `a_module_method': undefined method `fetch' for nil:NilClass (NoMethodError)
	from spec/failing.rb:31:in `<main>'
```

spec/module_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module ModuleMethods
```
joevandyk, line 3: this spec seems to pass

spec/failing.rb
```diff
@@ -0,0 +1,31 @@
+require 'memo_wise'
```
joevandyk, line 1: this program seems to fail

joevandyk:
I added a Dockerfile to the PR to eliminate any environment-specific issues that be happening on my machine..

```
➜  docker build -t memo_wise_script .
[+] Building 0.8s (8/8) FINISHED                                                                                                                                                                             docker:desktop-linux
 => [internal] load build definition from Dockerfile                                                                                                                                                                         0.0s
 => => transferring dockerfile: 132B                                                                                                                                                                                         0.0s
 => [internal] load .dockerignore                                                                                                                                                                                            0.0s
 => => transferring context: 2B                                                                                                                                                                                              0.0s
 => [internal] load metadata for docker.io/library/ruby:3.2                                                                                                                                                                  0.6s
 => [1/4] FROM docker.io/library/ruby:3.2@sha256:4298936e80cf36804f8c5e6bc0bb3539e78557d99bd54bf033263610c02e2fac                                                                                                            0.0s
 => [internal] load build context                                                                                                                                                                                            0.0s
 => => transferring context: 859B                                                                                                                                                                                            0.0s
 => CACHED [2/4] RUN gem install memo_wise                                                                                                                                                                                   0.0s
 => [3/4] COPY spec/failing.rb failing.rb                                                                                                                                                                                    0.0s
 => ERROR [4/4] RUN ruby failing.rb                                                                                                                                                                                          0.1s
------
 > [4/4] RUN ruby failing.rb:
0.136 /usr/local/bundle/gems/memo_wise-1.8.0/lib/memo_wise.rb:184:in `a_module_method': undefined method `fetch' for nil:NilClass (NoMethodError)
0.136 	from failing.rb:30:in `<main>'
0.136 ClassWithoutInitializer
------
Dockerfile:7
--------------------
   5 |     COPY spec/failing.rb failing.rb
   6 |
   7 | >>> RUN ruby failing.rb
   8 |
--------------------
ERROR: failed to solve: process "/bin/sh -c ruby failing.rb" did not complete successfully: exit code: 1
```

joevandyk:
@JacobEvelyn The tests probably aren't formatted exactly as they should be, but I think it illustrates the problem.

joevandyk:
@JacobEvelyn any thoughts on this one?

joevandyk:
Thanks for looking at this!

To be honest, I don't know why the instance variable isn't always initialized. This fix is a bandaid for what I assume is properly initializing the variable, unfortunately my Ruby's not good enough to understand this gem at that deep a level. :)

JacobEvelyn:
Okay I dug in a bit more here. The challenge is that even if we change our `def memo_wise` to have something like this:

```ruby
def klass.included(base)
  base.class_eval <<~HEREDOC, __FILE__, __LINE__ + 1
    def initialize(#{ALL_ARGS})
      MemoWise::InternalAPI.create_memo_wise_state!(self)
      super
    end
  HEREDOC
end
```

this only works if the class we care about defines `initialize` _before_ `include`ing the module:

```ruby
class MyClass
  def initialize(*) ; end
  include MyModule
end
```

This is because the `included` method gets called immediately and the `initialize` method it adds will get overridden by the one in `MyClass` if the `include` happens first.

It's not ideal, but I wonder if a solution for your use case is to just use `prepend` for your module instead of `include`?

There's also some other work going on right now in #324 and #326 to fix some similar issues, and I'm trying to see if a similar solution would work here.

ms-ati:
What if we change MemoWise to actually error out if we use include instead of prepend?

JacobEvelyn:
> What if we change MemoWise to actually error out if we use include instead of prepend?

@ms-ati I think the issue is it's not caused by someone `include`ing MemoWise, they're `include`ing their own module which `prepend`s MemoWise.

@joevandyk I'm no longer confident I can get something working on the code to resolve this issue for you. Would one of these workarounds work for you instead?

- change your `include ModuleMethods` to `prepend ModuleMethods`
- move your `include ModuleMethods` to be below your `def initialize`?

ms-ati:
QQ: What would happen if everything stayed the same (that is: `include ModuleMethods` at the top, before the `def initialize`), but in `initialize` it calls `super` either at beginning or end?

Does that still trigger the MemoWise initialize?

ms-ati:
@JacobEvelyn I am wondering if we should consider making use of [Module#method_added](https://ruby-doc.org/3.2.2/Module.html#method-i-method_added) as a callback, and **if the method added is** `#initialize`, take some action?

sunny:
I’ve had issues with ActiveRecord’s latest marshalling format.

```rb
ActiveRecord.marshalling_format_version = 7.1

class User < ActiveRecord::Base
  prepend MemoWise
end

Rails.cache.fetch("test") { User.new }.instance_variable_get(:@_memo_wise)
# => {}
Rails.cache.fetch("test") { User.new }.instance_variable_get(:@_memo_wise)
# => nil
```

This branch fixes the issue 🙏🏻

JacobEvelyn:
Since all the issues that have been linked here have been resolved, I'm going to close this as well. I've noted @ms-ati's idea from [this comment](https://github.com/panorama-ed/memo_wise/pull/321#issuecomment-1908939505) on [this open issue](https://github.com/panorama-ed/memo_wise/issues/382#issuecomment-2856495199).

PR #327, opened 2024-01-30
Prepend ivar setup in included modules

JacobEvelyn:
**Before merging:**


lib/memo_wise.rb
```diff
@@ -94,6 +94,20 @@ def inherited(subclass)
   end
   private_constant(:CreateMemoWiseStateOnInherited)
 
+  module CreateMemoWiseStateOnIncluded
+    def included(base)
+      return unless base.is_a?(Class) && !base.singleton_class?
+
+      base.prepend(Module.new do
+        def initialize(...)
```
JacobEvelyn, line 102: I think this will need to change to support earlier Ruby versions, but just keeping it as-is for a proof-of-concept.

lib/memo_wise.rb
```diff
@@ -94,6 +96,15 @@ def inherited(subclass)
   end
   private_constant(:CreateMemoWiseStateOnInherited)
 
+  module CreateMemoWiseStateOnIncluded
+    def included(base)
```
JacobEvelyn, line 100: @alpaca-tc in your code example you had an additional conditional here and below. Am I missing the reasoning behind those?

ms-ati (approved):
LGTM

lib/memo_wise.rb
alpaca-tc: Basically, I don't want to define unnecessary methods, so I added a branch.

BTW, I found a bug that memo_wise is not initialized when a module is included in a class with an included module.
We can fix this case by simply making the following change, but there may be other edge cases.

Something may need to be done to reduce the complexity.　

```
diff --git a/spec/memo_wise_spec.rb b/spec/memo_wise_spec.rb
index 6e73d3a..a2355a1 100644
--- a/spec/memo_wise_spec.rb
+++ b/spec/memo_wise_spec.rb
@@ -365,18 +365,32 @@ RSpec.describe MemoWise do
           end
         end
 
+        let(:nested_module_with_initializer) do
+          Module.new do
+            include Module3
+            def initialize(*); end
+          end
+        end
+
         let(:klass_with_module_with_initializer) do
           Class.new do
             include Module3
           end
         end
 
+        let(:klass_with_nested_module_with_initializer) do
+          Class.new do
+            include Module4
+          end
+        end
+
         let(:instance) { klass.new }
 
         before(:each) do
           stub_const("Module1", module1)
           stub_const("Module2", module2)
           stub_const("Module3", module_with_initializer)
+          stub_const("Module4", nested_module_with_initializer)
         end
 
         it "memoizes inherited methods separately" do
@@ -402,6 +416,14 @@ RSpec.describe MemoWise do
           expect(instance.module1_method_counter).to eq(1)
         end
 
+        it "can memoize klass with nested module with initializer" do
+          instance = klass_with_nested_module_with_initializer.new(true)
+          expect { instance.module1_method }.not_to raise_error
+
+          expect(Array.new(4) { instance.module1_method }).to all eq("module1_method")
+          expect(instance.module1_method_counter).to eq(1)
+        end
+
         it "can reset klass with initializer" do
           instance = klass_with_initializer.new(true)
           expect { instance.reset_memo_wise }.not_to raise_error
@@ -411,6 +433,11 @@ RSpec.describe MemoWise do
           instance = klass_with_module_with_initializer.new(true)
           expect { instance.reset_memo_wise }.not_to raise_error
         end
+
+        it "can reset klass with nested module with initializer" do
+          instance = klass_with_nested_module_with_initializer.new(true)
+          expect { instance.reset_memo_wise }.not_to raise_error
+        end
       end
 
       context "when the class, its superclass, and its module all memoize methods" do
```

```
diff --git a/lib/memo_wise.rb b/lib/memo_wise.rb
index e793603..4a5ee2a 100644
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -97,10 +97,16 @@ module MemoWise
   private_constant(:CreateMemoWiseStateOnInherited)
 
   module CreateMemoWiseStateOnIncluded
+    module CreateMemoWiseStateOnInitialize
+      class_eval(INITIALIZE_LITERAL, __FILE__, __LINE__ + 1)
+    end
+
     def included(base)
-      base.prepend(Module.new do
-        class_eval(INITIALIZE_LITERAL, __FILE__, __LINE__ + 1)
-      end)
+      if base.is_a?(Class)
+        base.prepend(CreateMemoWiseStateOnInitialize)
+      elsif !base.singleton_class?
+        base.singleton_class.prepend(CreateMemoWiseStateOnIncluded)
+      end
     end
   end
   private_constant(:CreateMemoWiseStateOnIncluded)
```

joevandyk:
Any reason to not merge?

JacobEvelyn:
@joevandyk thanks for the nudge. Not sure what happened to this PR but I can take a look in the new year. Would love to get it merged!

joevandyk:
@JacobEvelyn awesome, thanks!

lib/memo_wise.rb
```diff
@@ -175,6 +183,8 @@ def memo_wise(method_name_or_hash)
         if klass.is_a?(Class) && !klass.singleton_class?
           klass.singleton_class.prepend(CreateMemoWiseStateOnInherited)
         else
+          klass.singleton_class.prepend(CreateMemoWiseStateOnIncluded) if klass.is_a?(Module) && !klass.singleton_class?
```
dburgess560, line 186: While I don't see the dual creation an issue, I'm wondering if having a combined CreateMemoWiseStateOnIncludedAndInherited would gain a slight performance bump. Any chance you took a peek at this? Not sure if difference between the two spots of klass.singleton_class.prepend and klass.prepend prevent this though

dburgess560 (approved):

lib/memo_wise.rb
dburgess560: Chatted live and updated some tests to confirm that we do indeed need the singleton module state created on include separately from inherited (else parent/child return the same cached content)

CHANGELOG.md
```diff
@@ -9,9 +9,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
-_No breaking changes!_
+- Allowed memoization when `include`d modules `prepend MemoWise` and define `initialize` [[#327]](https://github.com/panorama-ed/memo_wise/pull/327)
 
-**Project enhancements:** none
+_Breaking changes:_
+- Removed Ruby 2.7 (EOL) and 3.0 (EOL) support to allow enhancement noted above [[#327]](https://github.com/panorama-ed/memo_wise/pull/327)
+
+**Project enhancements:**
+
+- Updated official test coverage to support Ruby 3.3 [[#335](https://github.com/panorama-ed/memo_wise/pull/335)]
```
ms-ati, line 19: This changelog confused me, since it seems like we are adding 3.4 here?

spec/memo_wise_spec.rb
```diff
@@ -351,11 +351,32 @@ def module2_method
           end
         end
 
+        let(:klass_with_initializer) do
+          Class.new do
+            include Module1
+            def initialize(*); end
```
ms-ati, line 357: Curious why `*` instead of `...` -- apologies I really do not remember the difference, just curious

ms-ati (approved):
A couple genuinly curious questions up, but optional to respond, this LGTM as long as all NDS tests pass with this version

spec/memo_wise_spec.rb
```diff
@@ -59,10 +59,10 @@
 
       it "memoizes methods with keyword and double-splat arguments" do
         expect(Array.new(4) { target.with_keyword_and_double_splat_args(a: 1, b: 2, c: 3) }).
-          to all eq("with_keyword_and_double_splat_args: a=1, kwargs={:b=>2, :c=>3}")
+          to all eq("with_keyword_and_double_splat_args: a=1, kwargs=#{{ b: 2, c: 3 }}") # rubocop:disable Lint/LiteralInInterpolation
```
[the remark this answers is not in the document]
JacobEvelyn: Annoyingly, the latest JRuby and TruffleRuby print hashes with `{b:2, c:3}` syntax while MRI uses hashrockets.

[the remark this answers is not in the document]
ms-ati: Wow that's awesome, thanks for clarifying!

spec/memo_wise_spec.rb
JacobEvelyn: I couldn't find a reason to use arguments at all (and confirmed the test passes without them but fails without our code change) so I removed them.

ms-ati (approved):
LGTM!

ms-ati (approved):

PR #336, opened 2024-05-28
rm Ruby 2.4 support to address rexml CVE

jordannb:
**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,23 @@ follows a format inspired by [Keep a Changelog](https://keepachangelog.com/en/1.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/panorama-ed/memo_wise/compare/v1.8.0...HEAD)
+## [Unreleased](https://github.com/panorama-ed/memo_wise/compare/v1.9.0...HEAD)
+
+**Gem enhancements:**
+
+_No breaking changes!_
+
+**Project enhancements:**
+
+## [v1.9.0](https://github.com/panorama-ed/memo_wise/compare/v1.8.0...v1.9.0)
 
 **Gem enhancements:**
 
 - Fixed a bug that overwrote existing self.extended method definitions. [[#324]](https://github.com/panorama-ed/memo_wise/pull/314)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Fixed a bug that overwrote existing self.extended method definitions. [[#324]](https://github.com/panorama-ed/memo_wise/pull/314)
 - Fixed a bug that overwrote existing self.inherited method definitions. [[#325]](https://github.com/panorama-ed/memo_wise/pull/315)
+- Removed Ruby 2.4 (EOL) support to allow upgrading rexml dependency version from a version that includes a [CVE](https://www.ruby-lang.org/en/news/2024/05/16/dos-rexml-cve-2024-35176/) [[#336]](https://github.com/panorama-ed/memo_wise/pull/336)
 
 _No breaking changes!_
```
JacobEvelyn, line 16: I think this _is_ a breaking change, right?

jordannb: @JacobEvelyn Yes! I added the note here temporarily until I got the PR build working, and now with #317 merged, I believe it's working.

So now I have a question - We want to both remove support for 2.4, to remove "sub-dependency" on a CVE-containing gem, and we want to add support for 3.3 (possibly with [YJIT benchmark](https://github.com/panorama-ed/memo_wise/commit/d0f4515e9454fdd81aa97b4e0120aa36fa4b0c5a)). Typically, I wouldn't do those two things in the same PR, since they seem like different things. Is there a preferred way to include these changes in the same release but not the same PR?

JacobEvelyn: Sorry I missed this! I don't think we have a super defined process here so whatever form of "merge two PRs and at some point bump the `VERSION` constant" makes sense to you works for me. I think we can get these out in pretty quick succession so I'm not too worried. In one sense we wouldn't be _adding_ support for 3.3 since the gem already works with 3.3, we'd just be adding CI tests/benchmarks for it. So if we wanted to be really thorough we could add 3.3 first in one PR, and then merge this PR with a `VERSION` bump?

Also, as a side note, I wasn't quickly able to get YJIT benchmarks working. I think it'll be a nice thing to have eventually but it wasn't even clear to me that the official ruby container can use YJIT so I wouldn't worry about that for now.

JacobEvelyn (approved):

PR #339, opened 2024-06-04
⚗️ Perform experiments on gems

### LINKED ISSUE #338: Documentation: Add alt_memery & memoist3 to performance comparison table
You have `memery` in the table already, but it has been nearly dead for a long time.  `alt_memery` is a maintained [fork](https://github.com/AlexWayfer/alt_memery) I've been using in production for a while now, which has hundreds of commits on top of the old `memery` 1.5.0 release.

I'll work on adding `alt_memery` to the comparison table.

pboling:
- Adds alt_memery to benchmarking
- Adds memoist3 to benchmarking
- Fixes #338

**Before merging:**


pboling:
Fixed rubocop lint issues.

JacobEvelyn:
Thanks for making this PR, @pboling ! Since we now have multiple gems with the same class names, I'm wondering:

1. Can we switch the table to display the gem name (e.g. `alt_memery`) instead of the class name? I think that would be much more readable here.
2. Can we find a creative way to get around the limitation of running the same class names next to each other? I'd really like to reserve the 2.7 table for gems that _don't_ support 3, not just for less-maintained ones. Some possible approaches would be [renaming the constant](https://stackoverflow.com/a/8762740/1103543) (my preferred path) or [loading the code a different way](https://github.com/panorama-ed/memo_wise/blob/main/benchmarks/benchmarks.rb#L7-L23).

Does that make sense?

pboling:
Regarding 1, yes, that's an easy switch.  Regarding 2, I wondered the same.  My only idea was to add a `refinement` which aliases the module namespace to a new constant, and then using that refinement in a block closure / context.   I think it wouldn't work though, because within the gem all references would still be to the old namespace, and there would be live overloading of the namespace happening.

But... if one of the two colliding gems doesn't have an internal references to its own namespace it could work...

JacobEvelyn:
Did you see the first link I shared, to rename the constant? Wouldn't that be simple and also handle internal references?

pboling:
Thinking more about this, and I don't think it is a good idea to spend time figuring out how to get two gems with the namespace to be loaded at the same time for a benchmarking run.

1. The namespaces will collide immediately when required by bundler, unless we add `require: false`, and require them manually, which incurs complexity, maintenance, and overhead for little value.
2. We would have to rewrite the code of one of each of the colliding gems, in the same way that is already being done for the GitHub checkout of memo_wise.  This could be made to work, but modifying an installed gem is fraught with problems, and the complexity is, IMO, simply not worth it for this script.
3. A simpler solution would be to have two scripts, one running a "heat" of gem comparisons without any colliding namespaces, and the other running a different heat of gem comparisons with the gems left out of the first heat.

pboling:
But thinking about it again... I'll just give it a try, and see if it works, the same way as is done for the MemoWise GH checkout.

pboling:
@JacobEvelyn sorry about the delay.  I have a gem called [`gem_bench`](https://github.com/pboling/gem_bench), which was the perfect home for the copy & re-namespacing of a gem so that it can be [benchmarked against itself](https://ruby.social/@galtzo/113151678496062402), or a fork of itself, and it took me awhile to distill that and test it.  All of that complexity (copy and renamespace) is now black-boxed, and it still happens before the benchmarking part starts, so it won't affect results.

Looking forward to your feedback!

benchmarks/benchmarks.rb
```diff
@@ -1,41 +1,79 @@
...
-
-# This string is both used for temp filepaths necessary to separate the GitHub
-# version of MemoWise and the local version, and used for the reported results
 GITHUB_MAIN = "MemoWise_GitHubMain"
-
-# We download a the main branch of MemoWise on GitHub into a tmp directory to
-# compare against the local version when we run benchmarks
-Dir.mktmpdir do |directory|
-  Dir["#{github_memo_wise_path}/lib/**/*.rb"].each do |file|
-    Tempfile.open([File.basename(file)[0..-4], ".rb"], directory) do |tempfile|
```
pboling, line 17: Note that in GemBench, `Tempfile` would not work with some libraries, because it makes the filename unique, and thus incompatible with later `require "ny_gem/my_file"` statements.  As a result I had to switch to using `File`, but still nested inside of `Dir.mktmpdir` for automatic cleanup.  The copied and renamespaced gems are loaded into memory, but only after the copy and renamespace is complete, and then the source is deleted.  This allows much more complex gems to be loaded.

README.md
```diff
@@ -114,36 +114,37 @@ For more usage details, see our detailed [documentation](#documentation).
...
+|----------------------------|---------------------|--------------------|--------------------|--------------------|
+| `()` (none)                | 19.77x              | 2.41x              | 21.53x             | 3.08x              |
+| `(a)`                      | 16.41x              | 11.80x             | 17.03x             | 10.38x             |
+| `(a, b)`                   | 15.09x              | 11.17x             | 15.21x             | 9.96x              |
+| `(a:)`                     | 20.91x              | 17.19x             | 18.48x             | 15.66x             |
+| `(a:, b:)`                 | 20.48x              | 17.30x             | 18.74x             | 16.18x             |
+| `(a, b:)`                  | 19.17x              | 16.03x             | 16.93x             | 15.57x             |
+| `(a, *args)`               | 2.91x               | 2.08x              | 2.90x              | 1.81x              |
+| `(a:, **kwargs)`           | 2.61x               | 2.20x              | 2.61x              | 2.18x              |
+| `(a, *args, b:, **kwargs)` | 2.13x               | 1.79x              | 1.99x              | 1.71x              |
```
pboling, line 146: Markdown tables are intended to look good and be readable in plaintext!  My IDE (RubyMine) does this automatically.

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 3.3.5
```
pboling, line 1: Support for modern multi-tool version managers, like `asdf` and `mise`.  Major IDE's also support it.

README.md
JacobEvelyn: Cool! I'm totally happy to have these nicely formatted, but would you be willing to change our benchmarks code so that it spits out a formatted table like this? Happy to help if that doesn't make sense.

pboling: That will take me a while. My IDE does it for me automatically, so I have never needed code to do it.

pboling: Perhaps to get this merged quicker I should revert to no spaces?

pboling: Done!

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ _No breaking changes!_
 **Project enhancements:**
 
 - Updated official test coverage to support Ruby 3.3 [[#335](https://github.com/panorama-ed/memo_wise/pull/335)]
+- Added `alt_memery` and `memoist3` to benchmarks [#339](https://github.com/panorama-ed/memo_wise/pull/339)
```
JacobEvelyn, line 27: Nitpick: add an extra `[]` here to match the PR link styling of other items

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ _No breaking changes!_
...
+  - Updated benchmark results in `README.md` to Ruby 3.3.5
```
JacobEvelyn, line 28: Nitpick: please make this a top-level item with its own PR link (even though it's the same):

```markdown
- Updated benchmark results in `README.md` to Ruby 3.3.5 [[#339](https://github.com/panorama-ed/memo_wise/pull/339)]
```

CHANGELOG.md
```diff
@@ -55,8 +57,7 @@ _No breaking changes!_
 - Updated `Dry::Core` gem version to 1.0.0 in benchmarks [[#297]](https://github.com/panorama-ed/memo_wise/pull/297)
```
JacobEvelyn, line 57: Feel free to say no, but while I have you here I just noticed I'd messed up some of the changelog link stylings in the v1.9.0 and v1.8.0 sections (the second `]` should come after the `)`). Would you mind fixing those?

benchmarks/Gemfile
```diff
@@ -6,16 +6,27 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby ">= 2.7.7"
 
-gem "benchmark-ips", "2.13.0"
+gem "benchmark", "0.3.0"
```
JacobEvelyn, line 9: I might've missed it below—what do we need this for?

benchmarks/Gemfile
```diff
@@ -6,16 +6,27 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
...
+gem "benchmark-ips", "2.14.0"
```
JacobEvelyn, line 10: Thanks for bumping this! 😻

benchmarks/Gemfile
```diff
@@ -6,16 +6,27 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
...
+# 2. memery & alt_memery share the namespace Memery
+# 3. memoist & memoist3 share the namespace Memoist, and also share a load path for their version.rb files.
+# This means we must `require: false` in `benchmarks/Gemfile` all, or all but one, of each of these duplicates,
+#   or we take care to only load them in discrete Ruby versions,
+#   to avoid a namespace collision before re-namespacing duplicates
 if RUBY_VERSION > "3"
+  gem "alt_memery", "2.1.0", require: false
   gem "dry-core", "1.0.1"
-  gem "memery", "1.5.0"
+  gem "memery", "1.6.0"
```
JacobEvelyn, line 23: Thanks for bumping this! 😻

benchmarks/Gemfile
```diff
@@ -6,16 +6,27 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby ">= 2.7.7"
 
-gem "benchmark-ips", "2.13.0"
+gem "benchmark", "0.3.0"
+gem "benchmark-ips", "2.14.0"
+gem "gem_bench", "2.0.1"
 
+# NOTE: Regarding `require: false` below
+# 1. GitHub version of MemoWise and the local source of MemoWise, share a namespace
```
JacobEvelyn, line 14: Nitpick: I think the comma on this line should be removed.

benchmarks/benchmarks.rb
```diff
@@ -59,26 +104,35 @@ def run_gc
...
+benchmarked_gems = re_namespaced_gems.select(&:required?).map do |re_namespaced_gem|
+  BenchmarkGem.new(
+    re_namespaced_gem.as_klass,
+    re_namespaced_gem.metadata[:activation_code],
+    re_namespaced_gem.metadata[:memoization_method],
+    re_namespaced_gem.gem_name == "memo_wise" ? GITHUB_MAIN_BENCHMARK_NAME : re_namespaced_gem.gem_name,
+  )
+end
+benchmarked_gems.push(
+    BenchmarkGem.new(MemoWise, "prepend MemoWise", :memo_wise, LOCAL_BENCHMARK_NAME),
```
JacobEvelyn, line 126: Looks like RuboCop is complaining that these are indented too much.

README.md
```diff
@@ -114,36 +114,36 @@ For more usage details, see our detailed [documentation](#documentation).
...
-|`()` (none)|0.60x|3.17x|
-|`(a)`|1.01x|7.94x|
-|`(a, b)`|0.85x|6.38x|
-|`(a:)`|1.00x|11.78x|
-|`(a:, b:)`|0.88x|9.67x|
-|`(a, b:)`|0.83x|9.44x|
-|`(a, *args)`|0.67x|1.45x|
-|`(a:, **kwargs)`|0.68x|1.88x|
-|`(a, *args, b:, **kwargs)`|0.64x|1.29x|
+Results using Ruby 3.3.5:
```
JacobEvelyn, line 117: Once everything else in this PR is good, I'll send you the latest benchmark output from GitHub Actions to stick in here (in case you don't have access to see it yourself).

JacobEvelyn:
Just took a closer look—overall pretty good, just a few small things to tidy up. Thanks so much for working on this!

benchmarks/benchmarks.rb
```diff
@@ -1,41 +1,86 @@
...
+      "Memoist" => "MemoistOne"
+    },
+    metadata: {
+      activation_code: "extend MemoistOne",
+      memoization_method: :memoize,
+    },
+  ),
+].each(&:doff_and_don) # Copies, re-namespaces, and requires each gem.
+
+# Here we require memo_wise gem from the `pwd` source.
```
JacobEvelyn, line 62: This is kind of confusing to me because we don't actually use `pwd` here. Can we just say we require the local version?

benchmarks/benchmarks.rb
```diff
@@ -1,41 +1,86 @@
...
+# Separately, in the Gemfile, we download the main branch of MemoWise from GitHub.
+# We've already moved it into a tmp directory and re-namespaced it to allow benchmark comparison with local source.
```
JacobEvelyn, line 64: Since you're talking about two different versions, could you clarify that "it" here refers to the GitHub `main` one? Maybe something like:

```ruby
# We've already installed the `memo_wise` version on the `main` branch from GitHub in the
# Gemfile, and moved it into a tmp directory and re-namespaced it so it doesn't collide with
# the `MemoWise` constant. Now we require the local version of `memo_wise` to compare
# this branch against it.
```

benchmarks/benchmarks.rb
```diff
@@ -1,41 +1,86 @@
...
-%w[memery memoist memoized memoizer ddmemoize dry-core].
+# in the Gemfile.  Gems re-namespaced by GemBench::Jersey will have already been loaded by now.
+%w[memery memoized memoizer ddmemoize dry-core].
   each { |gem| require gem if Gem.loaded_specs.key?(gem) }
 
+# Some Gems Have Modules Which Need To Be Required Manually:
+# 1. `dry-core` => Memoizable
+{
+  # file_path => gem_name
+  "dry/core/memoizable" => "dry-core",
```
JacobEvelyn, line 76: This feels a bit like overkill to me if this is the only gem we have like this. Thoughts on changing it back to the old version?

pboling: It had included other gems / files during the convulsions it went through to get where it is, and it is definitely overkill now.

benchmarks/benchmarks.rb
pboling: Yeah, that would be more clear.

benchmarks/Gemfile
pboling: At some point I was getting a failure that made me think this would help, and the failure went away, but it may not have been relevant at all.

Just tried, and it works without it, and here are no direct references.  Removing.

benchmarks/benchmarks.rb
pboling: FWIW, the `..` in the require path is relative to `pwd` in Ruby, and that's where my head was at, but it's an opaque reference.

pboling:
@JacobEvelyn The CI build is awaiting approval!

benchmarks/benchmarks.rb
```diff
@@ -1,41 +1,81 @@
...
-      tempfile.rewind
-      require tempfile.path
-    end
-  end
-end
-
+GITHUB_MAIN_BENCHMARK_NAME = "memo_wise-github-main"
+LOCAL_BENCHMARK_NAME = "memo_wise-local"
+
+# 1. GitHub version of MemoWise and the local source of MemoWise, share a namespace
```
JacobEvelyn, line 12: Eek another comma to remove here 🙈

pboling: Fixed!

README.md
```diff
@@ -114,36 +114,36 @@ For more usage details, see our detailed [documentation](#documentation).
...
+
+|Method arguments|`alt_memery` (2.1.0)|`dry-core` (1.0.1)|`memery` (1.6.0)|`memoist3` (1.0.0)|
```
JacobEvelyn, line 119: Latest result from GitHub Actions:

```markdown
|Method arguments|`alt_memery` (2.1.0)|`dry-core`\* (1.0.1)|`memery` (1.6.0)|`memoist3` (1.0.0)|
|--|--|--|--|--|
|`()` (none)|11.84x|0.67x|3.10x|2.58x|
|`(a)`|9.50x|1.11x|3.78x|15.21x|
|`(a, b)`|7.67x|0.93x|3.00x|12.06x|
|`(a:)`|15.99x|1.16x|7.12x|21.32x|
|`(a:, b:)`|12.83x|0.91x|5.70x|21.20x|
|`(a, b:)`|12.95x|0.94x|5.72x|17.11x|
|`(a, *args)`|1.89x|0.70x|0.74x|2.91x|
|`(a:, **kwargs)`|2.81x|0.69x|1.19x|4.65x|
|`(a, *args, b:, **kwargs)`|1.66x|0.58x|0.81x|2.80x|
```

README.md
```diff
@@ -114,36 +114,36 @@ For more usage details, see our detailed [documentation](#documentation).
...
+|`(a)`|6.60x|0.84x|5.70x|13.84x|
+|`(a, b)`|5.02x|0.72x|4.15x|10.53x|
+|`(a:)`|9.83x|0.83x|5.94x|17.24x|
+|`(a:, b:)`|7.66x|0.70x|4.66x|16.24x|
+|`(a, b:)`|7.87x|0.70x|4.64x|13.38x|
+|`(a, *args)`|2.08x|0.78x|1.57x|3.58x|
+|`(a:, **kwargs)`|2.46x|0.59x|1.78x|6.84x|
+|`(a, *args, b:, **kwargs)`|1.67x|0.67x|1.12x|4.10x|
 
 \* `Dry::Core`
```
JacobEvelyn, line 131: Change `Dry::Core` to `dry-core` here

README.md
```diff
@@ -114,36 +114,36 @@ For more usage details, see our detailed [documentation](#documentation).
...
 [may cause incorrect behavior caused by hash collisions](https://github.com/dry-rb/dry-core/issues/63).
 
 Results using Ruby 2.7.8 (because these gems raise errors in Ruby 3.x):
 
-|Method arguments|`DDMemoize` (1.0.0)|`Memoist` (0.16.2)|`Memoized` (1.1.1)|`Memoizer` (1.0.3)|
+|Method arguments|`ddmemoize` (1.0.0)|`memoist` (0.16.2)|`memoized` (1.1.1)|`memoizer` (1.0.3)|
```
JacobEvelyn, line 136:
```markdown
|Method arguments|`ddmemoize` (1.0.0)|`memoist` (0.16.2)|`memoized` (1.1.1)|`memoizer` (1.0.3)|
|--|--|--|--|--|
|`()` (none)|24.14x|2.44x|23.84x|2.59x|
|`(a)`|22.16x|14.80x|20.70x|11.67x|
|`(a, b)`|19.39x|13.66x|18.03x|11.46x|
|`(a:)`|30.54x|23.68x|25.21x|21.20x|
|`(a:, b:)`|27.75x|22.59x|23.47x|20.65x|
|`(a, b:)`|26.72x|21.39x|21.73x|19.43x|
|`(a, *args)`|3.26x|2.31x|3.09x|1.93x|
|`(a:, **kwargs)`|2.87x|2.29x|2.51x|2.10x|
|`(a, *args, b:, **kwargs)`|2.23x|1.88x|1.97x|1.73x|
```

JacobEvelyn:
@pboling final comments in; if you could make those README changes and then squash everything into one commit with a descriptive commit message this will be good to merge!

pboling:
@JacobEvelyn Rebased on latest main, squashed to single commit.  Looks ready to me.

JacobEvelyn (approved):

JacobEvelyn:
Thanks so much for all the work here @pboling !

PR #344, opened 2024-09-16
Optimize Logo (non-breaking reduction of the production gem size)

tagliala:
Hello,

this is being done in the optics of reduce https://rubygems.org/gems/memo_wise size

This is the first, non-breaking change that should reduce the size from 173KB to 91KB

However, in the next PR, `logo.png` will be excluded from the production build

---

```
optipng -o7 logo.png
```

Output file size = 66167 bytes (95622 bytes = 59.10% decrease)

**Before merging:**


JacobEvelyn:
Thanks! This looks great!

PR #345, opened 2024-09-16
Reduce production gem size

tagliala:
Hello,

please check if this makes sense and if the production gem should also include other files

---

Include only files that are needed for the production gem.

Also avoids to use git to allow to build the package in an environment
that does not have git (on purpose).

### Before

173K

```
.dokaz
.github/PULL_REQUEST_TEMPLATE.md
.github/dependabot.yml
.github/workflows/dependency-review.yml
.github/workflows/main.yml
.gitignore
.rspec
.rubocop.yml
.ruby-version
.yardopts
CHANGELOG.md
CODE_OF_CONDUCT.md
Gemfile
Gemfile.lock
LICENSE.txt
README.md
Rakefile
benchmarks/Gemfile
benchmarks/benchmarks.rb
bin/console
bin/setup
lib/memo_wise.rb
lib/memo_wise/internal_api.rb
lib/memo_wise/version.rb
logo/logo.png
memo_wise.gemspec
```

### After

19K

```
CHANGELOG.md
LICENSE.txt
README.md
lib/memo_wise.rb
lib/memo_wise/internal_api.rb
lib/memo_wise/version.rb
```

**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 **Gem enhancements:**
 
+- Reduced unzipped gem size from 173KB to 19KB [[#345](https://github.com/panorama-ed/memo_wise/pull/345)]
+
 _No breaking changes!_
 
 **Project enhancements:**
--- a/memo_wise.gemspec
+++ b/memo_wise.gemspec
@@ -26,13 +26,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = Gem::Requirement.new(">= 2.5.0")
 
   # Specify which files should be added to the gem when it is released.
-  # The `git ls-files -z` loads the files in the RubyGem that have been added
-  # into git.
-  spec.files = Dir.chdir(File.expand_path(__dir__)) do
-    `git ls-files -z`.split("\x0").reject do |f|
-      f.match(%r{^(test|spec|features)/})
-    end
-  end
+  spec.files = Dir.glob("{CHANGELOG.md,LICENSE.txt,README.md,lib/**/*.rb}")
   spec.require_paths = ["lib"]
 
   spec.metadata = {


memo_wise.gemspec
```diff
@@ -26,13 +26,10 @@ Gem::Specification.new do |spec|
...
-  # The `git ls-files -z` loads the files in the RubyGem that have been added
-  # into git.
-  spec.files = Dir.chdir(File.expand_path(__dir__)) do
-    `git ls-files -z`.split("\x0").reject do |f|
-      f.match(%r{^(test|spec|features)/})
-    end
-  end
+  spec.files = Dir.glob(
+    '{CHANGELOG.md,LICENSE.txt,README.md,lib/**/*.rb}',
+    File::FNM_DOTMATCH
```
JacobEvelyn, line 31: Why do we need `FNM_DOTMATCH`? Wouldn't it be the same without it since none of these files begin with `.`?

tagliala: Thanks, it is not needed, it is also wrong for this use case. I've used it somewhere else to exclude directories (`.`, `..`) but this is not the case

memo_wise.gemspec
```diff
@@ -26,13 +26,7 @@ Gem::Specification.new do |spec|
...
 
   # Specify which files should be added to the gem when it is released.
-  # The `git ls-files -z` loads the files in the RubyGem that have been added
-  # into git.
-  spec.files = Dir.chdir(File.expand_path(__dir__)) do
-    `git ls-files -z`.split("\x0").reject do |f|
-      f.match(%r{^(test|spec|features)/})
-    end
-  end
+  spec.files = Dir.glob('{CHANGELOG.md,LICENSE.txt,README.md,lib/**/*.rb}')
```
JacobEvelyn, line 29: Looks like RuboCop wants these `'`s to be `"`s. Other than that I think this is good!

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 **Gem enhancements:**
 
+- Reduced production gem size by 9 times [[#345](https://github.com/panorama-ed/memo_wise/pull/345)]
```
JacobEvelyn, line 12: To be honest I always find statements like this confusing (would `reduced size by 1x` mean no reduction or a 100% reduction?). Would you be open to changing this to absolutes, e.g.:

```
Reduced unzipped gem size from 173KB to 19KB
```

or something similar?

tagliala: Much better, thanks

JacobEvelyn:
I actually have a question for you: https://github.com/panorama-ed/memo_wise/pull/346#discussion_r1763482262

JacobEvelyn:
@tagliala I forgot to mention this yesterday, but this improvement has been published to RubyGems as v1.10.0. Thanks!

tagliala:
I've seen that, downloaded, and tested, thanks!

Issue #349, opened 2024-09-18
Benchmarks: are memowise results reliable?


tagliala:
Hello,

I've checked the main branch `ac1231e` and changed the version number.

```diff
diff --git a/lib/memo_wise/version.rb b/lib/memo_wise/version.rb
index 428f25a..a4fb14d 100644
--- a/lib/memo_wise/version.rb
+++ b/lib/memo_wise/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module MemoWise
-  VERSION = "1.10.0"
+  VERSION = "1.10.0.dev"
 end
```

then I've tried to run the benchmarks

```
~/dev/memo_wise/benchmarks on ac1231e [!$]
$ bundle exec ruby benchmarks.rb
Skipping memoist (not loaded on 3.3.3)

Will BENCHMARK_GEMS:
	dry-core (1.0.1)
	memoist3 (1.0.0)
	memo_wise-local (1.10.0)
	alt_memery (2.1.0)
	memery (1.6.0)
	memo_wise-github-main (1.10.0)
ruby 3.3.3 (2024-06-12 revision f1c7b6f435) [arm64-darwin23]
```

I was expecting to see `1.10.0.dev` in memo_wise-local and most important I was expecting to. see 1.00 (or something like that) in the comparison between memowise local and memowise github main, but I'm getting results like:

|Method arguments|`memo_wise-github-main` (1.10.0)|
|--|--|
|`()` (none)|0.99x|
|`(a)`|1.02x|
|`(a, b)`|1.00x|
|`(a:)`|1.01x|
|`(a:, b:)`|1.00x|
|`(a, b:)`|1.01x|
|`(a, *args)`|0.97x|
|`(a:, **kwargs)`|0.86x|
|`(a, *args, b:, **kwargs)`|1.00x|

I'm asking because I'm getting different results  for `(a, *args)` and `(a:, **kwargs)` almost all times

My config:
- different ruby versions, from 2.7 to 3.3 (both arm and x64)
- macOS 15.0 (24A335)
- rvm 1.29.12-next (master)

JacobEvelyn:
Hmm good catch; I too am seeing strange behavior. @pboling I expect `puts MemoWise::VERSION` to not succeed if I put it right after the `doff_and_don` call, but it prints `1.10.0`. I similarly can't get it to print `1.10.0.dev` if I modify that locally.

cc @ms-ati

tagliala:
[...]

PR #350, opened 2024-09-19
Prefer require_relative for internal requires

tagliala:
As a side note, this change has been approved here by `@deivid-rodriguez` which is familiar with this issue: https://github.com/activeadmin/arbre/pull/622

---

`require_relative` is preferred over `require` for files within the same
project because it uses paths relative to the current file, making code
more portable and less dependent on the load path.

This change updates internal requires to use `require_relative` for
consistency, performance, and improved portability.

Ref:
- https://github.com/ruby/psych/pull/522
- https://github.com/ruby/logger/pull/20
- https://github.com/ruby/rdoc/pull/658
- https://github.com/panorama-ed/memo_wise/issues/349
- https://github.com/rubocop/rubocop/issues/8748

---

**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 **Gem enhancements:**
 
+- Changed internal `require`s to `require_relative` to make code less dependent on the load path [[#350](https://github.com/panorama-ed/memo_wise/pull/350)]
+
 _No breaking changes!_
 
 **Project enhancements:**
--- a/lib/memo_wise.rb
+++ b/lib/memo_wise.rb
@@ -2,8 +2,8 @@
 
 require "set" # Ruby < 3.2 does not load `set` by default.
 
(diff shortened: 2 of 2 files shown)


pboling (approved):
LGTM

JacobEvelyn (approved):

tagliala:
@JacobEvelyn done and force pushed in the same commit

As a side note, for some extreme use cases, this can be considered a breaking change

Ref: https://github.com/varvet/pundit/pull/831#issuecomment-2367444446

pboling:
It is a potential breaking change for a use case that isn't actually a feature or intended.  What he describes is a way to break the Ben on purpose with knowledge of the source, so that you can hack a library in a very strange way.  It is an interesting thought experiment. Several people have now responded to him on the RuboCop discussion.

IMO it would be extreme to consider this a breaking change, especially since it fixes a real use case, which is benchmarking as this gem does.

JacobEvelyn (approved):

JacobEvelyn:
Yes I agree with both of these that this should not be considered a breaking change.

PR #351, opened 2024-09-20
⬆️ gem_bench v2.0.5

pboling:
Fixes the `require_relative` > `require` issue discussed in #349 for the `gem_bench` gem itself.  Does not merit a mention in the changelog, IMO.

**Before merging:**

### DIFF
--- a/benchmarks/Gemfile
+++ b/benchmarks/Gemfile
@@ -7,7 +7,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby ">= 2.7.7"
 
 gem "benchmark-ips", "2.14.0"
-gem "gem_bench", "2.0.3"
+gem "gem_bench", "2.0.4"
 
 # NOTE: Regarding `require: false` below
 # 1. GitHub version of MemoWise and the local source of MemoWise share a namespace
--- a/benchmarks/Gemfile
+++ b/benchmarks/Gemfile
@@ -7,7 +7,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby ">= 2.7.7"
 
 gem "benchmark-ips", "2.14.0"
-gem "gem_bench", "2.0.4"
+gem "gem_bench", "2.0.5"
 
 # NOTE: Regarding `require: false` below
 # 1. GitHub version of MemoWise and the local source of MemoWise share a namespace


pboling:
It failed to run the benchmarks because I forgot to enter my MFA code to actually push the new gem version to RubyGems before I made this change. 😬 🤦 It will pass if run again, no other changes needed.  The gem is released now.

JacobEvelyn:
Thanks! I'll wait to re-run this until after #350 is merged.

JacobEvelyn:
@pboling based on https://github.com/panorama-ed/memo_wise/issues/349#issuecomment-2359406124 I expected that with this code, if I add `puts MemoWise::VERSION` directly after the `doff_and_don` call I'd see an error, but I do not. What am I misunderstanding?

pboling:
@JacobEvelyn as it happens, I have another gem, [`require_bench`](https://github.com/pboling/require_bench), which is for mencharking and debugging `require` and `load` statements... I'll see if I can figure out how `Memowise::VERSION` is available at that point.

pboling:
OK, I figured out exactly what is happening.

I'd like to demostrate the solution with yet another gem of mine, [`version_gem`](https://gitlab.com/oauth-xx/version_gem), which I built to avoid this exact problem, but it will take me a bit to get around to a PR for that.

To briefly summarize - `gem_bench` uses `bundler` to figure out what a gem's path is so it can copy it.  In order to do this `bundler` has to load the available "gemspecs".  Most gems use a terrible practice (IMO) of requiring their `version.rb` file into their `gemspec`, and this makes it impossible for that namespace to not be loaded if bundler has accessed the gem.

Note: I use this in every gem I've published in the last few years.  See: https://gitlab.com/oauth-xx/version_gem/#usage for how to set it up.  You'll note that it only removes the `VERSION` constant after grabbing a local variable pointer to the value it holds.  This means that in the way I've been using it, it does still leave behind namespace pollution.  That could be avoidable, perhaps by removing the namespace as well, but perhaps namespace pollution is actually good?  This is the first I've thought of it.

In any case, using that pattern you could clean up all of the constants that get defined there, and have an entirely clean gemspec.  Also, to follow the pattern, you actually don't need `version_gem` at all.  It includes a few other nice-to-haves that I just happen to want everywhere.

The `version_gem` library was extracted from the very popular `oauth2` library a few years ago.

JacobEvelyn:
Thanks for all the details, @pboling, and sorry it's taken me so long to get back to this.

After pushing a branch to GitHub that makes your proposed gemspec change, and then using that branch in benchmarks locally, I can finally confirm that `MemoWise::VERSION` is not defined directly after the `doff_and_don` call!

The only thing I'm wondering about is why we need to use `load` in the gemspec instead of just `require_relative` as before. It seems to work for me either way. Do you know? (I'll also ask in the other GitHub thread where you found this trick.)

pboling:
So there were two things the load pattern addressed.

1. Constant already defined warning.

Essentially, require / require_relative will track things they load, so a subsequent load of the same resource will not be double loaded, except that it gets confused by dissimilar paths, so that many things actually do get double loaded.  The way I've dealth with this is to use `load`, so it is never tracked, and a later require / require_relative will act normally.  But because I've loaded the file, I then have to remove the constant manually to avoid the warning that the constant was already defined.

If it works fine with require_relative, and doesn't trigger the warning about the redefinition of the constant when the library loads the version.rb file again due to a different path, that's great.  I haven't followed how that may have improved in Ruby.  Its possible my load pattern may not be needed anymore.

2. Simplecov tracking code coverage.

Not relevant in this use case.

pboling:
I'll rebase this on top of your latest changes once you've merged #370 ⭐

JacobEvelyn:
@pboling once you rebase this I'll re-run CI to get the latest benchmark results

JacobEvelyn:
@pboling sorry, I ended up just doing this in #352

PR #356, opened 2024-09-26
Configure Style/NumericPredicate for performance

tagliala:
Hello,

While the [Style/NumericPredicate](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/NumericPredicate) cop recommends using `.zero?`, this can lead to performance degradation in gems where zero checks are frequent and speed-critical; for such performance-oriented libraries, using `== 0` remains preferable despite not adhering to the style guide.

This will not lead to performance improvements for the current implementation of memo_wise, I'm rather submitting this as a "reminder"

---

Replace predicate methods like `v.zero?` with direct comparison `v == 0` for better performance.

While the current implementation may not show visible improvements, frequently called methods using `.zero?` instead of `== 0` can lead to significant performance losses.

Benchmarks across Ruby versions 2.5 to 3.3 consistently show comparisons  outperforming predicates:

```
Comparison (Ruby 2.6 to 3.2, v = 0):
              v == 0: 21479329.3 i/s
             v.zero?: 17979885.4 i/s - 1.19x  (± 0.00) slower

Comparison (Ruby 2.5 and 3.3, v = 0):
              v == 0: 23652215.7 i/s
             v.zero?: 21843174.0 i/s - 1.08x  (± 0.00) slower

Comparison (Ruby 2.5 to 3.3, v = 1):
              v == 0: 23227474.2 i/s
             v.zero?: 21675200.9 i/s - 1.07x  (± 0.00) slower
```

**Before merging:**

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,3 +15,6 @@ Style/DocumentDynamicEvalDefinition:
 
 Style/MultipleComparison:
   Enabled: false
+
+Style/NumericPredicate:
+  EnforcedStyle: comparison
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -291,7 +291,7 @@ end
     benchmark_json.sort_by! { |json| json["name"] }
 
     # Print headers based on the first benchmark_json
-    if i.zero?
+    if i == 0
       benchmark_headers = benchmark_json.map do |benchmark_gem|
         # Gem name is of the form:
         # "memoist (1.1.0): ()"
--- a/lib/memo_wise/internal_api.rb
+++ b/lib/memo_wise/internal_api.rb
@@ -44,7 +44,7 @@ module MemoWise
     #   - :double_splat (examples: `def foo(a: 1)`, `def foo(a:, **b)`)
     #   - :splat_and_double_splat (examples: `def foo(a=1, b: 2)`, `def foo(a=1, **b)`, `def foo(*a, **b)`)
     def self.method_arguments(method)
-      return NONE if method.arity.zero?
+      return NONE if method.arity == 0
 
       parameters = method.parameters.map(&:first)
 


ms-ati:
Hi @tagliala! First I want to thank you for what appears to be an awareness campaign to help the community.

Can you explain why we would make this change here, in these two lines of code? One is in benchmark set up, the other is in one-time instrumentation of an object with MemoWise.

In neither case, is it yet clear to me that our trade-off would be for performance at the cost of (arguable) readability. Is that trade-off clear to you here? If so, can you help explain and make it clear?

tagliala:
Hi,

> the other is in one-time instrumentation of an object with MemoWise.

In the specific case of memo wise, it is not going to provide any visible improvement, because as you mentioned it is a one time initialization. This is also stated in the opening post: "_This will not lead to performance improvements for the current implementation of memo_wise, I'm rather submitting this as a "reminder_"

> Is that trade-off clear to you here? If so, can you help explain and make it clear?

It is more than clear. The fact is that this cop is globally configured for readability, so it may lead to an usage of numeric predicates in contexts where they are not optimal, like loops. Again, this is not the case. I see `memo_wise` as a gem oriented to performance, so I proposed this change here in specific, but as a reminder and also to read other thoughts

JacobEvelyn:
Thanks for making this PR, @tagliala! I appreciate the thoughtful commit message and interest in making future gem changes more performant.

Personally, I'm indifferent to this change. I generally like having performant defaults, but I also struggle to see a world in which we would change this gem to include `.zero?` or `.one?` in a hot path _at all_, let alone have it slip through code review _and_ our benchmarks. I'm not opposed to merging this however, since you took the time to make it.

@ms-ati do you have a preference?

tagliala:
Welcome

>  I'm not opposed to merging this however, since you took the time to make it.

Thanks, and don't worry about the time.

I was also proposing to understand if that method could have made the difference in some circumstances, like large codebases, or if the new default may positively have impact on future development, but as far as I understand this is not a concern for memo_wise. And if it will ever be, it will be possible to make the change in the future

JacobEvelyn:
> I was also proposing to understand if that method could have made the difference in some circumstances, like large codebases

On this point specifically I think it won't make a noticeable difference. That `.zero?` will typically run at load time (when `memo_wise` is called), and based on your benchmark it seems like both `.zero?` and `== 0` can execute many billions of times per second. I can't see a codebase ever having close to that many `memo_wise` calls so I don't think this will matter.

JacobEvelyn:
I'm going to close this, but please let me know if anyone feels this should not be closed.

PR #370, opened 2024-12-23
Undefine VERSION in gemspec to avoid namespace pollution between multiple MemoWises

JacobEvelyn:
This commit updates `memo_wise.gemspec` to undefine the `MemoWise::VERSION` constant after it is read, as first suggested in
https://github.com/simplecov-ruby/simplecov/issues/557#issuecomment-825171399.

Before this change, accessing `MemoWise::VERSION` immediately after the `doff_and_don` call in `benchmarks.rb` gives the version of the gem pulled from GitHub, even though `doff_and_don` should mean nothing is in the `MemoWise` namespace. After this change, accessing `MemoWise::VERSION` there gives this error as we expect:

```
uninitialized constant MemoWise (NameError)
```

**Before merging:**

### DIFF
--- a/memo_wise.gemspec
+++ b/memo_wise.gemspec
@@ -1,13 +1,30 @@
 # frozen_string_literal: true
 
-require_relative "lib/memo_wise/version"
-
 Gem::Specification.new do |spec|
-  spec.name     = "memo_wise"
-  spec.version  = MemoWise::VERSION
-  spec.summary  = "The wise choice for Ruby memoization"
+  spec.name = "memo_wise"
+
+  # After loading the `VERSION` constant, save it to a variable to use in this
+  # gemspec and then undefine it. This allows benchmarks to compare multiple
+  # versions of MemoWise against each other without inadvertently sharing the
+  # same `VERSION` constant.
+  # NOTE: In future Ruby versions, we can avoid this complexity and the use of
+  # `GemBench` in `benchmarks.rb` by using namespaces. For more context, see:
+  # https://bugs.ruby-lang.org/issues/21311
+  # NOTE: It's important that we use `load` instead of `require_relative`
+  # because the latter only loads a file once, and so by undefining the
+  # `VERSION` constant we can make it difficult to access that value again
+  # later. For more context, see: https://github.com/panorama-ed/memo_wise/pull/370#issuecomment-2560268423
+  spec.version = Module.new.tap do |mod|
+    # NOTE: We fully qualify `Kernel` here because Rubygems defines its own
+    # `load` method, which is used by default when this code is executed by
+    # `gem build memo_wise.gemspec`. That `load` method does not support the
+    # optional "wrap" parameter we pass through as `mod`. For more context, see:
+    # https://github.com/simplecov-ruby/simplecov/issues/557#issuecomment-2630782358
+    Kernel.load("lib/memo_wise/version.rb", mod)
+  end::MemoWise::VERSION
+  spec.summary = "The wise choice for Ruby memoization"
   spec.homepage = "https://github.com/panorama-ed/memo_wise"
-  spec.license  = "MIT"
+  spec.license = "MIT"
 
   spec.authors = [
     "Panorama Education",
--- a/benchmarks/benchmarks.rb
+++ b/benchmarks/benchmarks.rb
@@ -15,6 +15,9 @@ LOCAL_BENCHMARK_NAME = "memo_wise-local"
 # This means we must `require: false` in `benchmarks/Gemfile` all, or all but one, of each of these duplicates,
 #   or we take care to only load them in discrete Ruby versions,
 #   to avoid a namespace collision before re-namespacing duplicates
+# NOTE: In future Ruby versions, we can avoid the use of `GemBench` and the
+# complexity of `spec.version` in `memo_wise.gemspec` by using namespaces. For
+# more context, see: https://bugs.ruby-lang.org/issues/21311
 re_namespaced_gems = [
   GemBench::Jersey.new(
     gem_name: "memo_wise",


pboling (approved):
Nice :)

pboling (changes requested):
Update...

Thinking about this more. This could introduce a bug where another require / require_relative will not have the effect of the laoding the file because it thinks the file has already been loaded, and it doesn't know that you've removed the constant... thus the VERSION constant may never get loaded.

I do think it should be `load` here.

pboling:
Basically, you want to avoid the "tracking" side effect of require because you're going to remove the VERSION., and thus want other requires to work normally in all circumstances.

JacobEvelyn:
> This could introduce a bug where another require / require_relative will not have the effect of the laoding the file because it thinks the file has already been loaded, and it doesn't know that you've removed the constant... thus the VERSION constant may never get loaded.

@pboling that's what I expected to happen, but I wasn't able to reproduce it locally. I suspect I had a bug in my setup though—are you able to reproduce the issue you described?

pboling:
Ooooh.  Looking into this more deeply...

`require_relative` does not cache the loaded file like `require`. This means that if you use `require_relative` multiple times in a single program, Ruby will reload the file each time, even if it was previously loaded.

(According to AI Llama 2.5 4B 🗡️ )

pboling (approved):
You should be fine with `require_relative`. 💯

pboling:
I *do* think there is a cognitive benefit to `load`, as it is very clearly an indication that you want the load to be *not require*.  It is a hard break from the require paradigm, where most people, including me, only think of the difference between require and require_relative, as a performance benefit by avoiding the load path, and not about the difference in tracking, which I'd forgotten about.

pboling:
And that's quite enough bike shedding from me.  👯

JacobEvelyn:
> require_relative does not cache the loaded file like require. This means that if you use require_relative multiple times in a single program, Ruby will reload the file each time, even if it was previously loaded.

This is false. I just tried using `require_relative` to load the same file twice, and the second time it returned `false` and did not load a new constant I added to the file in between `require_relative` calls. So I think the concern still is valid; I just would like to be able to reproduce it with `MemoWise` before switching to `load`. Does that make sense?

pboling:
Wow.  So AI lied to me.  :). Glad to know my original thinking was correct.  I now remember when operationalized the load paradigm for my gems I wanted to use `require_relative`, and tried to make that work, but I couldn't, but most of that experience is lost in the fog of time.  I think I can create an example to demostrate the issue...

JacobEvelyn:
For context, I tried reproducing the issue with this `Gemfile`:

```
# frozen_string_literal: true

source "https://rubygems.org"

git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby ">= 2.7.7"

gem "memo_wise", github: "panorama-ed/memo_wise", branch: "version-undefined"
```

and this test Ruby file:

```
require 'memo_wise'

puts MemoWise::VERSION
```

ch-panorama (approved):

pboling:
Yep.  Playing with this myself, and I think there are two paths.

First, the paranoid path, which is what I've started down, but haven't committed to:
```
# final param `true` wraps the loaded namespace in an anonymous module, amking the constant collision impossible.
# downside is it uses a $LOAD_PATH search, and the inherent performance penalty.
load("lib/gem_name/version.rb", true)
```

Second, the hopeful path:
```
# works fine, so long as the gem's other lib files, and other tools, always use require / require_relative to also load the version file.
# updise is it does not use the $LOAD_PATH search, so has a performance benefit.
require_relative "lib/gem_name/version"
```

I can't fault either choice.

JacobEvelyn:
@pboling can you explain what you mean by

> so long as the gem's other lib files, and other tools, always use require / require_relative to also load the version file.

I don't think I follow.

JacobEvelyn:
@pboling gotcha, so something like this?

```ruby
Module.new.tap { |mod| load("lib/memo_wise/version.rb", mod) }::MemoWise::VERSION
```

pboling:
Recent versions of Ruby are able to pass a module, instead of a simple true/false wrap param, ever since this PR: https://github.com/ruby/ruby/pull/4986
Unfortunately most of my libraries support older versions of Ruby still, so that's not useful for me yet.  Prior to that feature, the only way to get access to the anonymous module that wrapped was a complex throw catch that straddled both the loading and the loaded files.

https://users.aalto.fi/~tontti/posts/ruby-by-examples/ruby-catching-anonymous-module/

So, my load idea isn't viable.  I suppose there isn't a good reason to not use `require_relative`, and I have a lot of code I should update.

Thanks for this discussion!

pboling:
@JacobEvelyn only saw your response after my previous reply, but yes.  That will work for gems that target recent versions of Ruby only.

memo_wise.gemspec
```diff
@@ -1,10 +1,19 @@
...
+# After loading the VERSION constant, save it to a variable to use in this
+# gemspec and then undefine it. This allows benchmarks to compare multiple
+# versions of MemoWise against each other without inadvertently sharing the same
+# VERSION constant.
+# NOTE: Other uses of this trick often switch `require_relative` to `load`, as
+# in https://github.com/simplecov-ruby/simplecov/issues/557#issuecomment-825171399,
+# but for our purposes `require_relative` seems to work fine. If any issues
+# arise, feel free to use `load` instead.
 require_relative "lib/memo_wise/version"
+gem_version = MemoWise.send(:remove_const, :VERSION)
```
pboling, line 12: So the problem now is if using require_relative, and then removing the constant, the file will not be reloaded later, and the constant will remain removed.  I'm not sure how this could work @JacobEvelyn

pboling (changes requested):

memo_wise.gemspec
pboling: I guess I'm full circle, to it needs to use `load`, and can optionally use the `wrap` param if on a new enough version of Ruby.

pboling:
Thinking about this more... still.  You made this PR so that you could isolate the versions of libraries in the same namespace.

Your solution is to remove the constant.  This works, however, the constant remains removed, and the version.rb file will not be reloaded by the gem.  This means that `MemoWise::VERSION` should remain undefined later.

Your [example in this comment](https://github.com/panorama-ed/memo_wise/pull/370#issuecomment-2560107388) is working but only because when you load the gem with `require 'memo_wise'` it doesn't load the gemspec... I think.

But I expect that there are scenarios where the gemspec and the gem could both be loaded, and in those scenarios, it could break, by not having a VERSION anymore, if something tried to reference it.

pboling:
Aha!

https://bundler.io/guides/bundler_in_a_single_file_ruby_script.html

The example for Bundler inline is actually exactly this scenario!

Below I've only changed it to reference MemoWise, but otherwise it is the same.

```
require 'bundler/inline'

gemfile do
  source "https://rubygems.org"
  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
  gem "memo_wise", github: "panorama-ed/memo_wise", branch: "version-undefined"
end

puts 'Gems installed and loaded!'
puts "MemoWise Version is #{MemoWise::VERSION}"
```

The result of running this:
```
❯ ruby main.rb
Gems installed and loaded!
main.rb:10:in `<main>': uninitialized constant MemoWise::VERSION (NameError)

puts "MemoWise Version is #{MemoWise::VERSION}"
```

If you use `load` instead, it will work 👍

JacobEvelyn:
Hmm @pboling when I try your code above (with `load`) I get:

```
/Users/jevelyn/.gem/ruby/3.3.3/gems/bundler-2.4.6/lib/bundler.rb:635:in `rescue in eval_gemspec':  (Bundler::GemspecError)
[!] There was an error while loading `memo_wise.gemspec`: cannot load such file -- lib/memo_wise/version. Bundler cannot continue.

 #  from /Users/jevelyn/.gem/ruby/3.3.3/bundler/gems/memo_wise-b06f1c6964e3/memo_wise.gemspec:17
 #  -------------------------------------------
 #  #   end::MemoWise::VERSION
 >  load "lib/memo_wise/version"
 #  gem_version = MemoWise.send(:remove_const, :VERSION)
```

pboling (approved):
LGTM

pboling:
Ah!  Sorry, typo.  load requires the file extension (very unexpected, but true!)

memo_wise.gemspec
```diff
@@ -1,10 +1,25 @@
...
+# the latter only loads a file once, and so by undefining the `VERSION` constant
+# we can make it difficult to access that value again later. For more context,
+# see: https://github.com/panorama-ed/memo_wise/pull/370#issuecomment-2560268423
+# NOTE: If we ever bump the minimum Ruby version to 3.1+, we can simplify this
+# code and use this instead:
+#
+#   spec.version = Module.new.tap do |mod|
+#     load("lib/memo_wise/version.rb", mod)
+#   end::MemoWise::VERSION
+load "lib/memo_wise/version"
```
pboling, line 17:
```suggestion
load "lib/memo_wise/version.rb"
```

pboling (changes requested):

pboling (approved):
LGTM

ch-panorama (approved):

pboling (approved):
LGTM

pboling:
Turns out that it should be `Kernel.load`, not `load`, because Rubygems defines a totally different load method that shadows the kernel one, and so what we thought was happening is not what was happening, even though it worked accidentally.
https://github.com/simplecov-ruby/simplecov/issues/557#issuecomment-2630782358

pboling (changes requested):

memo_wise.gemspec
```diff
@@ -1,13 +1,27 @@
...
+  # `VERSION` constant we can make it difficult to access that value again
+  # later. For more context, see: https://github.com/panorama-ed/memo_wise/pull/370#issuecomment-2560268423
+  spec.version = Module.new.tap do |mod|
+    # NOTE: We fully qualify `Kernel` here because Rubygems defines its own
+    # `load` method, which is used by default when this code is executed by
+    # `gem build memo_wise.gemspec`. That `load` method does not support the
+    # optional "wrap" parameter we pass through as `mod`. For more context, see:
+    # https://github.com/simplecov-ruby/simplecov/issues/557#issuecomment-2630782358
+    Kernel.load("lib/memo_wise/version.rb", mod)
+  end::MemoWise::VERSION
```
pboling, line 21: This is so cool. It's working right? :heart_decoration: I'm going to switch to use this pattern everywhere!

JacobEvelyn: Yep, works for me!

ch-panorama (approved):

ch-panorama (approved):

memo_wise.gemspec
JacobEvelyn: Also note [these comments](https://github.com/panorama-ed/memo_wise/compare/14bb4355e6fe96c7285d4ca643022acfbcd2eb9d..403a4e75d5ef9806e1bf7daa4c33fc0788365765) I just added

Issue #382, opened 2025-05-05
Allow memoization for instances whose classes include a module that includes a module prepending MemoWise


JacobEvelyn:
As noted by @alpaca-tc in [this comment](https://github.com/panorama-ed/memo_wise/pull/327#discussion_r1472837528), if we add these tests they fail:

```
diff --git a/spec/memo_wise_spec.rb b/spec/memo_wise_spec.rb
index 6e73d3a..a2355a1 100644
--- a/spec/memo_wise_spec.rb
+++ b/spec/memo_wise_spec.rb
@@ -365,18 +365,32 @@ RSpec.describe MemoWise do
           end
         end
 
+        let(:nested_module_with_initializer) do
+          Module.new do
+            include Module3
+            def initialize(*); end
+          end
+        end
+
         let(:klass_with_module_with_initializer) do
           Class.new do
             include Module3
           end
         end
 
+        let(:klass_with_nested_module_with_initializer) do
+          Class.new do
+            include Module4
+          end
+        end
+
         let(:instance) { klass.new }
 
         before(:each) do
           stub_const("Module1", module1)
           stub_const("Module2", module2)
           stub_const("Module3", module_with_initializer)
+          stub_const("Module4", nested_module_with_initializer)
         end
 
         it "memoizes inherited methods separately" do
@@ -402,6 +416,14 @@ RSpec.describe MemoWise do
           expect(instance.module1_method_counter).to eq(1)
         end
 
+        it "can memoize klass with nested module with initializer" do
+          instance = klass_with_nested_module_with_initializer.new(true)
+          expect { instance.module1_method }.not_to raise_error
+
+          expect(Array.new(4) { instance.module1_method }).to all eq("module1_method")
+          expect(instance.module1_method_counter).to eq(1)
+        end
+
         it "can reset klass with initializer" do
           instance = klass_with_initializer.new(true)
           expect { instance.reset_memo_wise }.not_to raise_error
@@ -411,6 +433,11 @@ RSpec.describe MemoWise do
           instance = klass_with_module_with_initializer.new(true)
           expect { instance.reset_memo_wise }.not_to raise_error
         end
+
+        it "can reset klass with nested module with initializer" do
+          instance = klass_with_nested_module_with_initializer.new(true)
+          expect { instance.reset_memo_wise }.not_to raise_error
+        end
       end
 
       context "when the class, its superclass, and its module all memoize methods" do
```

[...]

PR #392, opened 2025-05-31
Add proper memo_wise initialization for inherited classes

Galathius:
Hopefully it solves the issue from [this comment](https://github.com/panorama-ed/memo_wise/issues/382#issuecomment-2925136130).

This change should allow such usage:
```ruby
class BaseClass
  prepend MemoWise
end

class A < BaseClass
  def initialize; end # <--- important part, there is no `super`

  memo_wise def slow
    sleep 2
    puts 'slow result'
  end
end
```

Now it's failing with:
```
/Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/memo_wise-1.13.0/lib/memo_wise.rb:174:in `slow': undefined method `fetch' for nil:NilClass (NoMethodError)
	from memo_wise_test.rb:26:in `block in <main>'
	from /Users/galathius/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/benchmark-0.4.1/lib/benchmark.rb:305:in `measure'
	from memo_wise_test.rb:26:in `<main>'
```

**Before merging:**


README.md
```diff
@@ -116,17 +116,17 @@ Benchmarks are run in GitHub Actions, and the tables below are updated with ever
 
 Results using Ruby 3.4.3:
 
-|Method arguments|`alt_memery` (2.1.0)|`dry-core`\* (1.1.0)|`memery` (1.7.0)|`memoist3` (1.0.0)|`short_circu_it` (0.29.3)|
```
Galathius, line 119: |Method arguments|`memo_wise-github-main` (1.13.0)|
|--|--|
|`()` (none)|1.00x|
|`(a)`|1.00x|
|`(a, b)`|0.98x|
|`(a:)`|0.99x|
|`(a:, b:)`|1.01x|
|`(a, b:)`|1.00x|
|`(a, *args)`|1.05x|
|`(a:, **kwargs)`|0.97x|
|`(a, *args, b:, **kwargs)`|1.00x|

JacobEvelyn:
Thanks for the PR, @Galathius ! I'm not sure this is right, though—your tests pass even when I remove the `def inherited` you added, and it appears that `CreateMemoWiseStateOnInherited` exists to provide this functionality but clearly isn't working correctly. I'll dig and see if I can find a more robust approach.

Edit: I actually don't understand the issue you're encountering; see https://github.com/panorama-ed/memo_wise/issues/382#issuecomment-2931036360

Galathius:
Hi @JacobEvelyn! Sorry, seems I accidentally removed `def initialize; end` from the child class in tests. 

Just updated it, now if you comment my code in `memo_wise.rb` you'll see it's failing.

PR #397, opened 2025-08-10
Fix loading gem specification externally

pboling:
When a gem's specification is loaded by another gem, to get at the metadata, or other info, the pathing for loading the `version.rb` won't work properly. This isn't an issue when running the gem on its own, but other tools that do analysis of other libraries will not be able to load this gemspec. Fortunately the fix is very simple, just prefix `#{__dir__}/`, and `__dir__` was introduced in Ruby 1.9.1, so shouldn't pose a problem.

I discovered this as I am now building a gem that loads other gem's gemspecs to get the data, with the goal of helping them fund-raise. https://github.com/galtzo-floss/floss_funding

**Before merging:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/panorama-ed/memo_wise/compare/v1.13.0...HEAD)
 
-**Gem enhancements:** none
+**Gem enhancements:**
+
+- Fixed issue with external tools loading `memo_wise.gemspec` from different directories [[#397]](https://github.com/panorama-ed/memo_wise/pull/397))
 
 _No breaking changes!_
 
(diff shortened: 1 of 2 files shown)


JacobEvelyn:
Thanks @pboling , and sorry I missed this! Do you mind adding a `CHANGELOG.md` entry for this? Something like:

```markdown
## [Unreleased](https://github.com/panorama-ed/memo_wise/compare/v1.13.0...HEAD)

**Gem enhancements:**

- Fixed issue with external tools loading `memo_wise.gemspec` from different directories [[#397]](https://github.com/panorama-ed/memo_wise/pull/397))

_No breaking changes!_

**Project enhancements:** none
```

pboling:
Will do. I think I may have improved the pattern a bit since this PR. Will review.

pboling:
@JacobEvelyn Rebase is done! Added to CHANGELOG.md.

pboling:
@JacobEvelyn ping!

